PR #41, opened 2020-10-26
Added RecordId to case response

LBHBReynolds-Carr:


SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
```diff
@@ -78,6 +78,8 @@ private static CareCaseData ToResponse(this FormData formData, BsonDocument rawD
         {
             return new CareCaseData
             {
+                Timestamp = DateTime.Now,
```
LBHMGeorgieva, line 81: This should be retrieving the timestamp from the form data

LBHBomotoso (approved):
All good

PR #49, opened 2020-11-11
Add asc allocations end point

LBHTKarki:

This adds the API endpoint for searching allocations for adults

New end point. No changes to existing functionality

Add endpoint details to Swagger hub doc


SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -116,6 +118,18 @@ public IActionResult GetChildrensAllocatedWorker([FromQuery] ListAllocationsRequ
             return Ok(_childrenAllocationUseCase.Execute(request));
         }
 
+        /// <summary>
+        /// Find acs allocations by Mosaic ID or officer email
+        /// </summary>
+        /// <response code="200">Success. Returns allocations related to the specified ID or officer email</response>
+        [ProducesResponseType(typeof(AscAllocationList), StatusCodes.Status200OK)]
+        [HttpGet]
+        [Route("asc_allocations")]
```
LBHMGeorgieva, line 127: All looks good, please just change the route to asc-allocations, instead of using underscore. I also updated the other endpoint

LBHMGeorgieva (approved):
Looks good. Please see comment regarding route

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
LBHTKarki: I've updated the route

PR #62, opened 2020-11-26
Implemented mosaic ID search

LBHBReynolds-Carr:


SocialCareCaseViewerApi/V1/Gateways/ProcessDataGateway.cs
```diff
@@ -31,24 +31,39 @@ public ProcessDataGateway(ISccvDbContext sccvDbContext, ISccvDbContextTemp sccvD
...
-                Builders<BsonDocument>.Filter.Regex("case_note_type", BsonRegularExpression.Create(new Regex(request.CaseNoteType, RegexOptions.IgnoreCase))) : null;
+            List<BsonDocument> result;
+            long mosaicId = 0;
+            if (!string.IsNullOrEmpty(request.MosaicId))
+            {
+                if (!Int64.TryParse(request.MosaicId, out mosaicId))
+                {
+                    throw new Exception();
+                }
+                var filter = "{$or: [{ mosaic_id: " + mosaicId.ToString() + "}, { mosaic_id: /" + mosaicId.ToString() + "/}]}";
```
LBHTKarki, line 42: is that 'or' required in the filter if we are setting it against single property?

PR #64, opened 2020-11-30
Integrated exact_name_match flag

LBHBReynolds-Carr:
- Default is false
- If true if statement triggers an exact match regex
- Both exact_match and fuzzy search are case insensitive
- Removed 'cases-test' route from controller
- Removed SccvDbContextTemp
- Removed unnecessary overload in usecase and gateway


SocialCareCaseViewerApi/V1/Gateways/ProcessDataGateway.cs
```diff
@@ -47,10 +47,20 @@ public IEnumerable<CareCaseData> GetProcessData(ListCasesRequest request)
             }
             else
             {
-                var firstNameFilter = !string.IsNullOrWhiteSpace(request.FirstName) ?
-                    Builders<BsonDocument>.Filter.Regex("first_name", BsonRegularExpression.Create(new Regex(request.FirstName, RegexOptions.IgnoreCase))) : null;
-                var lastNameFilter = !string.IsNullOrWhiteSpace(request.LastName) ?
-                    Builders<BsonDocument>.Filter.Regex("last_name", BsonRegularExpression.Create(new Regex(request.LastName, RegexOptions.IgnoreCase))) : null;
+                if (!request.ExactNameMatch)
+                {
+                    firstNameFilter = !string.IsNullOrWhiteSpace(request.FirstName) ?
```
LBHTKarki, line 52: we are matching names separately so it's not full name exact match but I think that's OK for now

LBHTKarki (approved):
Looks OK

PR #112, opened 2021-01-29
Allocation of worker in system

LBHBReynolds-Carr:
- Added FormName to CaseNoteBase
- Added CreatedBy and AllocationId to DeallocationBaseNote
- Added CreatedBy and AllocationId to AllocationBaseNote
- Added relevant parameters to Create and Update Allocation requests
- Set FormName in database gateway


SocialCareCaseViewerApi/V1/Boundary/Requests/CreateAllocationRequest.cs
```diff
@@ -14,5 +14,11 @@ public class CreateAllocationRequest
 
         [Required]
         public string AllocatedBy { get; set; }
```
LBHTKarki, line 16: I don't think this is necessary anymore, but we can sort it out when we update this to support teams as well. We get the allocated worker id anyway, so just need the CreatedBy as additional value

LBHTKarki (approved):

PR #114, opened 2021-02-04
Person ordering

LBHBReynolds-Carr:
- Removed obsolete test (contained in a separate commit)
- Switch function created for all criteria
- Ternary functions within criteria to sort by ascending or descending
- Default is sort by dateOfEvent, order by desc
- Function tested locally


LBHTKarki:
I've resolved an issue with default timestamp/date of event ordering in this same PR. I also removed the additional default date ordering which was left over from the previous implementation

PR #130, opened 2021-03-01
Improved create case notes endpoint

LBHTKarki:
Updated case note endpoint with validation. Notes now inline with imported data


LBHTKarki:
Updated create case note end point. Also merged hotfix branches

PR #149, opened 2021-03-23
Post warning notes endpoint

LBHBReynolds-Carr:
SCS-547 (https://hackney.atlassian.net/browse/SCS-547)

- New infrastructure objects:
-- Warning Note Set
-- Warning Note Case Note

- New Boundary objects:
-- CreateWarningNoteRequest
-- CreateWarningNoteResponse

- CreateWarningNote method created in DatabaseGateway
- WarningNoteUseCase object created with ExecutePost method
- warningnotes route created in controller
- CreateWarningNote method created in SCCVAPIController


SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -299,5 +303,26 @@ public IActionResult ListVisits([FromQuery] ListVisitsRequest request)
...
+        /// Create warning note.
+        /// <response code="201">Record successfully created</response>
+        /// <reponse code="500">There was a problem creating the record</reponse>
+        /// </summary>
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [HttpPatch]
+        [Route("warningnotes")]
+        public IActionResult CreateWarningNote([FromQuery] CreateWarningNoteRequest request)
+        {
+            try
```
LBHSPreston, line 317: What about other types of exceptions?  How are these handled?  Do you want these to fail hard or gracefully?

SocialCareCaseViewerApi.Tests/V1/Boundary/Response/CreateWarningNoteResponseTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        private CreateWarningNoteResponse _classUnderTest;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _classUnderTest = new CreateWarningNoteResponse();
+        }
+
+        [Test]
+        public void ResponseHasWarningNoteId()
```
LBHSPreston, line 19: Maybe just semantics but the names of these test methods don't reflect what they are testing.  A possible way of testing an entity property might be `Assert.That(entity, Has.Property("WarningNoteId").InstanceOf(typeof(int)));`

LBHBReynolds-Carr: I'll flag that with Tuomo, as if we're changing it here it might be best to also change all the boundary and infrastructure tests that have the same format, which could be done as a separate branch

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
LBHBReynolds-Carr: Currently the CreateWarningNoteException is the only one that I've built that might be thrown, could you recommend any other exceptions that I should be expecting?

LBHSPreston: Looking at the program flow from the controller to the gateway for that endpoint and the only exception handling I saw was where the CreateWarningNoteException was caught.  As there is an external resource dependency (gateway) where you have less control, it might be best to handle any possible issues accessing that external resource with a try/catch block.  This will allow the application to fail more gracefully and give you more control on what happens if that dependency fails.

LBHSPreston (approved):

PR #152, opened 2021-03-24
Show historic data based on feature flag value

Renny4Real:
JIRA: https://hackney.atlassian.net/browse/SCS-595

We want to be able to toggle showing historic case notes and visits while we test the feature with the users.

Use the feature flag value set in AWS as an environment variable to determine whether or not code to retrieve historic case note or historic visit data is executed.

#### _Checklist_

- Test feature flag set up works in staging
- Clean up feature flag implementation after production release of historic mosaic data

Any feedback/suggestions would be appreciated 🙂

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
@@ -270,23 +270,53 @@ namespace SocialCareCaseViewerApi.Tests.V1.Controllers
         }
 
         [Test]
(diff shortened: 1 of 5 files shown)


chubberlisk (approved):
Commits and tests were really easy to follow so it was easy to review! 🙌 I'm not 100% sure a `500` response code is what we should return but it's easy enough to change so 🚢 .

Renny4Real:
@chubberlisk Yeah 😅 . Just spoke with @neilkidd and he helped clear up a confusion I had.

Going to change them to `200`s.

So that the frontend doesn't need to worry about handling extra stuff that would have been normal functionality. 

I got confused and thought that perhaps we wanted to return an error response when we spoke about returning a response when the feature flag is set to false.

chubberlisk (approved):

PR #153, opened 2021-03-24
Update GetCaseNoteById API endpoint to return 404s

chubberlisk:

Currently, the API endpoint: `api/v1/casenotes/{id}` returns a `500` response when the platform API can't find a case note with the provided ID instead of a `404`.

This PR updates `SocialCareCaseViewerApiController#GetCaseNoteById` to catch the `SocialCarePlatformApiException` and when it's a `404` return `404 Not Found`.

## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCS-514


Renny4Real (approved):
🎉

PR #159, opened 2021-03-29
Update staging postgres RDS engine version

Renny4Real:
- Terraform errors with this message: "Cannot find version 11. for Postgres" when the engine version is set as "11."
- Updating the engine version to match the chosen version for the production Postgres RDS instance to achieve parity.


chubberlisk (approved):
🚢

PR #161, opened 2021-03-30
Add test data

Renny4Real:
* Using  [Mockaroo](https://www.mockaroo.com), we generated some SQL statements to add test data that could be used locally or in staging.

* Also removed tabs in the schema file as these caused errors when running the SQL statements.

Tables populated:
- dm_persons
- dm_addresses
- dm_telephone_numbers
- sccv_worker
- sccv_team
- sccv_workerteam


neilkidd (approved):
LGTM
Could you add the tables populated in the PR description please. It means later on we won't need to double check the scripts.

database/seeds.sql
```diff
@@ -0,0 +1,70 @@
...
+-- add test teams
+insert into dbo.sccv_team (id, name, context) values (32, 'Rutrum', 'C');
+insert into dbo.sccv_team (id, name, context) values (15, 'Sodales', 'A');
+insert into dbo.sccv_team (id, name, context) values (74, 'Primis', 'C');
+insert into dbo.sccv_team (id, name, context) values (35, 'Aenean', 'C');
+insert into dbo.sccv_team (id, name, context) values (20, 'Tristique', 'C');
+insert into dbo.sccv_team (id, name, context) values (24, 'Lectus', 'A');
+
+-- add test workers
+insert into dbo.sccv_worker (id, email, first_name, last_name, role) values (21, 'zmctague0@eventbrite.com', 'Zebedee', 'McTague', 'est');
```
LBHTKarki, line 49: Should we update the emails to use some fake domains, so it doesn't look like we are promoting something? These will also be visible in the app so might raise some questions there.

LBHTKarki: Ideally we should use dummy domains. If we update this source and then update the existing records in the db, I think that should work. Then it would be fine moving forward.

neilkidd: [http://example.com/](http://example.com/) is a good candidate?

chubberlisk: Yup, definitely! We'll change to all to `@example.com`.

chubberlisk: Updated in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/161/commits/cd4e39cfe7a04f930c5a3b7d7764ffb5af236990. We'll update the staging DB afterwards.

PR #165, opened 2021-04-01
Update CircleCI and serverless config to deploy to Mosaic-Production

chubberlisk:

We want to move the service API to the Mosaic-Production AWS account as that is more restrictive than the Production-APIs account that it currently lives in.

To do this first though, we want to deploy the service API into both ProductionAPIs and Mosaic-Production at the same time and then when everything set up correctly and test, we can change where frontend points to and "switch off" the service API in ProductionAPIs.

This PR updates the CircleCI and serverless config to deploy to Mosaic-Production as a separate workflow.

What this means is when `development` is merged into `master`, CircleCI will deploy the service API in Mosaic-Production in parallel with deploying to ProductionAPIs.

### *Follow-up actions after merging PR*

We'll need to merge `development` into `master` (but need to check in with team about this - if not can create the S3 bucket needed in the meantime) and then make sure things have been configured correctly.

## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCS-608

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,21 @@ commands:
           root: *workspace_root
           paths:
             - .aws
+  assume-role-and-persist-workspace-mosaic-production:
+    description: "Assumes deployment role and persists credentials across jobs"
+    parameters:
+      aws-account:
+        type: string
+    steps:
+      - checkout
+      - aws_assume_role/assume_role:
+          account: <<parameters.aws-account>>
+          profile_name: default
+          role: "circleci-assume-role"
+      - persist_to_workspace:
+          root: *workspace_root
+          paths:
+            - .aws
   terraform-init-then-apply:
     description: "Initializes and applies terraform configuration"
     parameters:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -38,6 +38,21 @@ commands:
           root: *workspace_root
           paths:
             - .aws
+  assume-role-and-persist-workspace-mosaic-production:
+    description: "Assumes deployment role and persists credentials across jobs"
```
Renny4Real, line 42: I think this uses the same description as the other `assume-role-and-persist-workspace` command. Wondering if it's worth mentioning which jobs these credentials would be persisted across 🤔

Renny4Real (approved):
Looks good

.circleci/config.yml
chubberlisk: Ah yeah, I like that idea. I'll update to `Assumes deployment role and persists credentials across jobs for Mosaic-Production`.

chubberlisk: Updated in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/165/commits/189a02d8dda8cb793ed64de994649484bc18a48c.

Renny4Real (approved):
🙌🏾

PR #167, opened 2021-04-07
Bring Database in line with front end requirements

LBHBReynolds-Carr:
- Added missing columns to DB Schema
- Added missing properties to Infrastructure and Request
- Changed property names to match with FE expectations
- Renamed all instances of CreateWarningNote to PostWarningNote


LBHTKarki (approved):
We can remove those unnecessary property tests later so you can leave everything as is for now

PR #168, opened 2021-04-07
Fix include team data and worker allocations

JohnFarrellDev:

/workers endpoint returns inconsistent data:
search by email gets back allocationCount, search by team_id I get back 1 (that is the correct value)
teams is always null even if I am worker is part of a team

Include allocations and worker team details when getting worker by email

Set team information to be included with response

#### _Checklist_

Staging check


sirLisko (approved):
The code looks good, didn't test it thought 👍

PR #173, opened 2021-04-09
Fixaroo serverless config for Mosaic-Production

chubberlisk:

After deploying to Mosaic-Production, it was noticed that the environment variables for the Lambda were not decrypted which meant that the values weren't as we expected and it had the security group for the platform API RDS instance...

This PR updates the serverless config to decrypt the SSM parameters needed and adds the correct security group IDs for Mosaic-Production.

Continue to fix and check. 🕵️‍♀️ 

## Link to JIRA ticket

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -24,12 +24,12 @@ functions:
     role: lambdaExecutionRole
     timeout: 30
     environment:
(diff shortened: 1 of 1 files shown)


Renny4Real (approved):
🎉

neilkidd (approved):

PR #174, opened 2021-04-12
Fix incorrect subnets for Mosaic-Production in serverless config

chubberlisk:

When testing the endpoint for getting a case note (i.e. testing the connection between the platform API and the service API), the service API in Mosaic-Production times out.

Looking at the VPC config for the service API and platform API, I noticed that the subnets for both were different. After making the subnets that the service API are in the same as the platform API and testing the endpoint all was good! ✅  

This PR updates the subnets for the service API to be the same as the platform API which are in the private subnets in Mosaic-Production.

Update the platform API's serverless config as this is defined incorrectly and not the same as what it is in AWS.

## Link to JIRA ticket

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -157,8 +157,8 @@ custom:
         - sg-0d8739383f0960389
         - sg-0d92164e9f5d53800
       subnetIds:
-        - subnet-0665104ee973a21be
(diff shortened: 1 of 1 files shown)


Renny4Real (approved):
👍🏾

PR #177, opened 2021-04-14
Refactor social care platform api gateway tests

JohnFarrellDev:
## Link to JIRA ticket

No associated jira ticket, just saw an option to refactor after previous work for historic visits that had me write tests for social care gateway api.

Tests were a bit messy/hard to read with many repeated code, entities created in helper functions now and mock the http client in a helper function.

#### _Checklist_

Potentially use as a good example of how to refactor.

Saw potential for refactor in previous visit ticket but didn't want to implement code outside the scope of the ticket as it would create a large PR.
Instead of waiting for a "grand refactor" one day in the future decided to tidy up after the previous visit PR was made.


SocialCareCaseViewerApi.Tests/V1/Gateways/SocialCarePlatformAPIGatewayTests.cs
```diff
@@ -16,288 +21,130 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
...
-            _httpClient = new HttpClient(mockHttpMessageHandler.Object)
-            {
-                BaseAddress = _mockBaseUri
-            };
-
-            _socialCarePlatformAPIGateway = new SocialCarePlatformAPIGateway(_httpClient);
+            var caseNote1 = TestHelper.CreateCaseNote();
+            var caseNote2 = TestHelper.CreateCaseNote();
+            var caseNotes = new ListCaseNotesResponse { CaseNotes = new List<CaseNote> { caseNote1, caseNote2 } };
+            var httpClient = CreateHttpClient(caseNotes);
```
LBHTKarki, line 33: Saw this set up in the other PR 👍

SocialCareCaseViewerApi.Tests/V1/Gateways/SocialCarePlatformAPIGatewayTests.cs
```diff
@@ -307,30 +154,22 @@ public void GivenHttpClientReturnsBadRequestResponseThenGatewayThrowsSocialCareP
...
+            return _httpClient;
         }
 
-
-        [Test]
-        [TestCase(HttpStatusCode.InternalServerError)]
-        [TestCase(HttpStatusCode.BadRequest)]
-        [TestCase(HttpStatusCode.NotFound)]
-        public void GivenHttpClientReturnsNon200ResponseThenGatewayThrowsSocialCarePlatformApiExceptionWithStatusCode(HttpStatusCode code)
+        private HttpClient CreateHttpClient<T>(T content)
```
LBHTKarki, line 170: 👍

SocialCareCaseViewerApi/V1/Domain/Visit.cs
```diff
@@ -1,14 +1,37 @@
...
 {
     public class Visit
     {
-        public string Id { get; set; }
-        public DateTime CreatedOn { get; set; }
-        public string CreatedByEmail { get; set; }
-        public string MosaicId { get; set; }
-        public string Title { get; set; }
-        public string Content { get; set; }
+        public long VisitId { get; set; }
```
LBHTKarki, line 7: I take it this has been validated with FE devs so it has all the properties they need?

SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
```diff
@@ -80,24 +80,16 @@ public static List<BsonDocument> HistoricalCaseNotesToDomain(List<CaseNote> case
...
-            }
-
-            return historicalVisits;
+            return visits.Select(visit => new BsonDocument(new List<BsonElement>
+                {
+                    new BsonElement("_id", visit.VisitId),
+                    new BsonElement("worker_email", visit.CreatedByEmail),
+                    new BsonElement("form_name_overall", "Historical_Visit"),
+                    new BsonElement("form_name", "Historical Visit"),
+                    new BsonElement("timestamp", visit.ActualDateTime ?? visit.PlannedDateTime),
```
LBHTKarki, line 89: Dates have been a bit of a headache because we are importing data as well and that can have different formats. They mainly use "dd/MM/yyyy H:mm:ss" format, so just need to check that search filters etc. still work OK. We have a ticket for normalising data across the board.

SocialCareCaseViewerApi.Tests/V1/Domain/VisitTests.cs
```diff
@@ -1,54 +0,0 @@
-using System;
-using NUnit.Framework;
-using SocialCareCaseViewerApi.V1.Domain;
-
-namespace SocialCareCaseViewerApi.Tests.V1.Domain
-{
-    public class VisitTests
```
LBHTKarki, line 7: Should we stop writing these sort of tests altogether? As long as there is proper test coverage these are probably pointless. We have discussed checking for properties and their types in objects, but that might not provide any additional value.

SocialCareCaseViewerApi/V1/Domain/Visit.cs
JohnFarrellDev: I've shown Luca the data being returned. 

For the /cases endpoint I will still get the Visit class to transform into what the FE needs (for when showing a list of records). 

For getting a specific Visit the frontend should want all the Visit information so plan is to return it all in the format from the new Visit class. I will confirm.

SocialCareCaseViewerApi.Tests/V1/Domain/VisitTests.cs
JohnFarrellDev: I don't believe there to be any value in these tests.

I don't think there is any value in testing for the properties either.

SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
JohnFarrellDev: The datetime coming from platform API is in the iso-8601 format as a string which is what we tend to aim for.

PR #180, opened 2021-04-15
Add component diagram

chubberlisk:

We're looking to improve documentation and visuals of the different systems that make up the Social Care system. [Social Care System Architecture](https://github.com/LBHackney-IT/social-care-architecture/tree/main) provides the high-level view and for lower level diagrams, these will live within the individual repos but linking back to the architecture repo.

This PR adds a component diagram for the service API. It's been created by using PlantUML

### Still to do

- [ ] Fix exported diagram being cropped off.
- [ ] Add guidance for how to update the diagram has also been added to the README.
- [ ] Update descriptions of repos in [container diagram](https://github.com/LBHackney-IT/social-care-architecture/blob/main/images/system-container.iuml).


neilkidd:
Superceded by #193

PR #181, opened 2021-04-15
Update some environment variables to be decrypted in serverless config

chubberlisk:

After deploying to Mosaic-Production, it was noticed that the environment variables for the MongoDB Lambda were not decrypted which meant that the values weren't as we expected and the Lambda can't connect to the MongoDB.

This PR updates the serverless config to decrypt the SSM parameters needed for the MongoDB Import Lambda in Mosaic-Production.

Double check environment variables values in Mosaic-Production and check import works.

## Link to JIRA ticket

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -53,12 +53,12 @@ functions:
     package:
       artifact: ./MongoDBImport/bin/release/netcoreapp3.1/mongodb-import.zip
     environment:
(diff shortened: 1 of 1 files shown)


neilkidd (approved):
LGTM - have I just got de-ja-vue with this PR 🤔

chubberlisk:
> LGTM - have I just got de-ja-vue with this PR 🤔

You have... I fixed it for the service API Lambda but not the MongoDB Lambda... 😅

PR #193, opened 2021-04-16
Add component diagram

neilkidd:
### What is the problem we're trying to solve

We're looking to improve documentation and visuals of the different systems that make up the Social Care system. Social Care System Architecture provides the high-level view and for lower level diagrams, these will live within the individual repos but linking back to the architecture repo.

### What changes have we introduced

This PR adds a component diagram for the service API. It's been created by using PlantUML

### Still to do
 
- [x] Add guidance for how to update the diagram has also been added to the README.
- [x]  Update descriptions of repos in container diagram in the architecture repo.


chubberlisk (approved):
💯

PR #202, opened 2021-04-19
Fix historic case note response for records history and viewing case note details

chubberlisk:

In the records history for a person, the “Record type” or `form_name` or “form type” returned for historic case notes is set to the `title` of it. This makes it difficult to filter by type on the frontend because the type for a historic case note is not generic.

This PR resolves the above by updating the `ResponseFactory#HistoricalCaseNotesToDomain` and `ResponseFactory#ToResponse` for a `CaseNote` which ensures that for records history i.e. `api/v1/cases`:

- we return `NoteType` for `form_name` to allow for filtering by type
- we continue to return `Title` via added `title` attribute

and for both records history and view historic case note by ID i.e. `api/v1/casenotes/{id}`:

- if `NoteType` is null, we return `"Case note"`
- if `NoteType` includes brackets e.g. `"Manager's Decisions (YH)"`, we return `"Manager's Decisions"`

#### _Checklist_

Notify frontend of changes.

## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCS-708

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
@@ -97,8 +97,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         [Test]
         public void HistoricalCaseNotesToDomainReturnsFormNameWithoutBracketsAndWhitespaceWhenNoteTypeHasASCInBrackets()
         {
-            var historicalCaseNote = TestHelpers.CreateCaseNote();
-            historicalCaseNote.NoteType = "Case Summary (ASC)";
+            var historicalCaseNote = TestHelpers.CreateCaseNote(noteType: "Case Summary (ASC)");
 
             var result = ResponseFactory.HistoricalCaseNotesToDomain(historicalCaseNote);
 
@@ -108,8 +107,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         [Test]
         public void HistoricalCaseNotesToDomainReturnsFormNameWithoutBracketsAndWhitespaceWhenNoteTypeHasYOTInBrackets()
         {
-            var historicalCaseNote = TestHelpers.CreateCaseNote();
-            historicalCaseNote.NoteType = "Home Visit (YOT)";
+            var historicalCaseNote = TestHelpers.CreateCaseNote(noteType: "Home Visit (YOT)");
 
             var result = ResponseFactory.HistoricalCaseNotesToDomain(historicalCaseNote);
 
@@ -119,8 +117,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         [Test]
         public void HistoricalCaseNotesToDomainReturnsFormNameWithoutBracketsAndWhitespaceWhenNoteTypeHasYHInBrackets()
         {
-            var historicalCaseNote = TestHelpers.CreateCaseNote();
-            historicalCaseNote.NoteType = "Manager's Decisions (YH)";
+            var historicalCaseNote = TestHelpers.CreateCaseNote(noteType: "Manager's Decisions (YH)");
 
             var result = ResponseFactory.HistoricalCaseNotesToDomain(historicalCaseNote);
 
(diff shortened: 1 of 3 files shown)


SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
```diff
@@ -117,5 +117,14 @@ public static WorkerResponse ToResponse(this Domain.Worker worker)
                 Teams = worker.Teams
             };
         }
+
+        private static string FormatFormNameForHistoricCaseNote(string noteType)
+        {
+            string pattern = @"\([^()]*\)$"; // Match brackets at the end e.g. (ASC)
+            var formName = noteType ?? "Case note";
```
JohnFarrellDev, line 124: We might want to use `String.IsNullOrEmpty` here instead of nullish coalescing just in case there are any empty strings ("")

JohnFarrellDev (approved):
There is a single change to make but apart from that I believe this is good. You can merge in straight after without another review.

Just make sure to make the frontend is aware and see that it is working in staging then we can go to prod.

SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
```diff
@@ -82,6 +83,50 @@ public void CanMapCaseNoteToBsonDocument()
...
+            var result = ResponseFactory.HistoricalCaseNotesToDomain(historicalCaseNote);
+
+            result.GetValue("form_name").AsString.Should().BeEquivalentTo("Case note");
+        }
+
+        [Test]
+        public void HistoricalCaseNotesToDomainReturnsFormNameWithoutBracketsAndWhitespaceWhenNoteTypeHasASCInBrackets()
+        {
+            var historicalCaseNote = TestHelpers.CreateCaseNote();
+            historicalCaseNote.NoteType = "Case Summary (ASC)";
```
JohnFarrellDev, line 101: We could update the TestHelpers.CreateCaseNote to use Random.String2 for the note type but also accept the value as a parameter for when we want to explicitly set it.

SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
chubberlisk: That makes sense, will do dat!

SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
chubberlisk: That would be good, however I'm not sure how to then set the expectation without having the test code the same as prod code. Any ideas how we could get around that? 🤔

chubberlisk: Ahhh never mind I got it, so we set the test helper to have the note types to select from without the brackets and then for these tests I can explicitly set it through a parameter.

JohnFarrellDev: I don't think the test code will be doing the exact same as prod.

In this example we'd have

```
[Test]
public void HistoricalCaseNotesToDomainReturnsFormNameWithoutBracketsAndWhitespaceWhenNoteTypeHasASCInBrackets()
{
    var historicalCaseNote = TestHelpers.CreateCaseNote(noteType: "Case Summary (ASC)");

    var result = ResponseFactory.HistoricalCaseNotesToDomain(historicalCaseNote);

    result.GetValue("form_name").AsString.Should().BeEquivalentTo("Case Summary");
}
```

Fundamentally test is doing the same thing we just aren't mutating the data post TestHelper function.

chubberlisk: Ahh noicee I like that - will update thanks!

SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
chubberlisk: Updated in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/202/commits/0311598b3a21b7c440b20d8a7232897dc295f77f.

SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
chubberlisk: Updated in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/202/commits/7bdaa07e3d8b4c5c11f90a57f784a87ed601ccd2.

PR #203, opened 2021-04-19
Update make commands and README for testing

chubberlisk:
This PR:

- ensures we're calling the right service for the `test` make command
- adds `start-test-dbs` make command to run the test databases in the background / detached mode
- updates the README with the above updates for testing and how to filter tests if you want to only run a selective few

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ shell:
 
 .PHONY: test
 test:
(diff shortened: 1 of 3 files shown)


JohnFarrellDev (approved):
This is really helpful, thanks and approved!

PR #204, opened 2021-04-19
Removal of feature flags historic visits and allocation dates

JohnFarrellDev:
## Link to JIRA ticket

No relevant jira ticket, this is cleanup after historic visits and allocations have gone into production.

We want to be proactive with out feature flags and remove them when they are no longer required.

Removal of feature flags and conditionals that required them.

#### _Checklist_

Deploy to prod (merge into master and deploy on circle ci) once we are confident staging is working as expected for the updated features.

### DIFF
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/CreateAllocationRequest.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/CreateAllocationRequest.cs
@@ -19,7 +19,7 @@ namespace SocialCareCaseViewerApi.V1.Boundary.Requests
         public string CreatedBy { get; set; }
 
         [JsonPropertyName("allocationStartDate")]
-        public DateTime? AllocationStartDate { get; set; }
+        public DateTime AllocationStartDate { get; set; }
     }
 
     public class CreateAllocationRequestValidator : AbstractValidator<CreateAllocationRequest>
@@ -38,6 +38,8 @@ namespace SocialCareCaseViewerApi.V1.Boundary.Requests
             RuleFor(x => x.CreatedBy)
                 .NotNull().WithMessage("Email Required")
                 .EmailAddress().WithMessage("Enter a valid email address");
+            RuleFor(x => x.AllocationStartDate)
+                .NotNull().WithMessage("Allocation start date required");
         }
     }
 }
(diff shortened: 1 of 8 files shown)


terraform/production/feature_flags.tf
```diff
@@ -1,8 +0,0 @@
-resource "aws_secretsmanager_secret" "show_historic_data_feature_flag" {
```
chubberlisk, line 1: Just a heads up incase you're not aware, this will have to be removed manually through the AWS console as this was never run. Maybe we delete this file for production but keep for staging as that does work to avoid confusion?

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateAllocationRequest.cs
```diff
@@ -38,6 +38,8 @@ public CreateAllocationRequestValidator()
             RuleFor(x => x.CreatedBy)
                 .NotNull().WithMessage("Email Required")
                 .EmailAddress().WithMessage("Enter a valid email address");
+            RuleFor(x => x.AllocationStartDate)
+                .NotNull().WithMessage("Allocation start date required");
```
chubberlisk, line 42: I think we're missing a test like `CreateAllocationReturns400WhenInvalidCreatedBy` but for `AllocationStartDate`. 🕵️‍♀️

terraform/production/feature_flags.tf
chubberlisk: We'll also need to remove the environment variable from the serverless config:

https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/e1b935cac6e76ab8c38cf43c0024d47029e526f2/serverless.yml#L38

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateAllocationRequest.cs
JohnFarrellDev: I agree this test is nice, problem I have to implement is that CreateAllocationRequest datetime property of AllocationStartDate is not nullable.

This is fine because we validate it exists and it means in the rest of the code we don't require any null checks and can be certain we have access to the property AllocationStartDate.

When I try to make a test with a null AllocationStartDate the compiler will not allow it :(

Arguably because of strict typing we don't need the test but I don't like that argument in this context.

terraform/production/feature_flags.tf
chubberlisk: Ah nice, yeah that makes sense!

chubberlisk (approved):

PR #206, opened 2021-04-20
Fix GET warning notes

Renny4Real:
## Link to JIRA ticket
https://hackney.atlassian.net/browse/SCS-717

## Describe this PR

When the GET `warningnotes/{id}` is called, the API returns a 500.

From investigating the logs, the error returned by the lambda read:
`Unknown error responding to request: InvalidCastException: System.InvalidCastException: Column is null`

It looks like Entity Framework returns this error when we've set up columns as required in the infrastructure when the actual database schema has the columns as NULLABLE.

This removes the extra `[required]` tags on columns that are nullable according to the database schema from within the Warning Notes Infrastructure Class

#### _Checklist_

Refactor the GET warning notes endpoint to match a more RESTful API endpoint structure 
i.e `GET /residents/{personId}/warningNotes`

### DIFF
--- a/SocialCareCaseViewerApi/V1/Infrastructure/WarningNote.cs
+++ b/SocialCareCaseViewerApi/V1/Infrastructure/WarningNote.cs
@@ -16,46 +16,49 @@ namespace SocialCareCaseViewerApi.V1.Infrastructure
         public long PersonId { get; set; }
 
         [Column("start_date")]
-        [Required]
         public DateTime? StartDate { get; set; }
 
         [Column("review_date")]
-        [Required]
         public DateTime? ReviewDate { get; set; }
 
         [Column("end_date")]
         public DateTime? EndDate { get; set; }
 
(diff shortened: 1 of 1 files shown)


1JW1 (commented):
Looks good. The code has been tidied, the changes to 'required' amended and maxlength removed, I presume, to provide more room for the actor to enter details.

Renny4Real:
> Looks good. The code has been tidied, the changes to 'required' amended and maxlength removed, I presume, to provide more room for the actor to enter details.

Yeah, primarily this was to have the infrastructure file match the schema of the database. 
So where the schema defines the column as `NOT NULL`, we've left the `[required]` tags and removed where it is not needed.

And wherever the schema defines the column to have a max length of n, we've added a `[MaxLength(n)]`.
The schema file can be found [here](https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/master/database/schema.sql).

PR #207, opened 2021-04-20
Refactor response factory tests

JohnFarrellDev:

I saw an opportunity for a minor refactor to tidy up some tests and create useful TestHelpers.
I saw this oppurtunity whilst working on a different but not to scope creep I decided to do the piece of work seperately. 

Added some TestHelpers and removed large amounts of hard coded test data.
In my opinion this makes our testing more robust but also makes tests easier to read/understand.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
@@ -1,10 +1,14 @@
 using System;
+using System.Collections.Generic;
 using Bogus;
 using Bogus.DataSets;
 using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Boundary.Response;
 using SocialCareCaseViewerApi.V1.Domain;
 using SocialCareCaseViewerApi.V1.Infrastructure;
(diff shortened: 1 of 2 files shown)


SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
```diff
@@ -46,6 +50,19 @@ public static CaseNote CreateCaseNote()
                 .RuleFor(c => c.CreatedByName, f => f.Person.FirstName);
         }
 
+        public static CaseNoteResponse CreateCaseNoteResponse(CaseNote? caseNote)
+        {
+            return new Faker<CaseNoteResponse>()
+                .RuleFor(c => c.RecordId, f => caseNote?.CaseNoteId ?? f.Lorem.Paragraph())
```
chubberlisk, line 56: I think this should be `f.UniqueIndex.ToString()`?

chubberlisk (approved):
satifised_seal

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
JohnFarrellDev: Thank you! Nice catch :)

PR #208, opened 2021-04-20
Feature/edit person

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/jira/software/projects/SCS/boards/51?selectedIssue=SCS-578

This update completes the edit person feature by providing two new endpoints, GET and PATCH.

Create person is already in place, so this will enable users to edit person records as well.

Two new end points for fetching and updating person details.

This feature will be refactored and moved over to platform API soon, so this is just a temporary function to tie us over until then. I wanted to do all the database updates with just one transaction, so the gateway has more logic than I would like, but that can be reviewed once the architecture for the new platform API is in place.

I started calling use cases etc. Person rather than resident to separate these out from the existing functionality. Person is more appropriate name for it from backend point of view.

#### _Checklist_

- [ x] Added tests to cover all new production code
- [ x] Code pipeline builds correctly

Hand the feature over for user testing


SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
```diff
@@ -117,5 +118,35 @@ public static WorkerResponse ToResponse(this Domain.Worker worker)
                 Teams = worker.Teams
             };
         }
+
+        public static GetPersonResponse ToResponse(Person person)
```
JohnFarrellDev, line 132: Do we want to be more explicit with the compiler about what can and can't be null.

If we use `#nullable enable` [C# Docs](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) we can then be explicit if a Type can be also null. For instance we then know person definitely is not null.

Currently we could pass null to this function and would null reference error on line 129 for example.

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -102,6 +104,50 @@ public IActionResult AddNewResident([FromBody] AddNewResidentRequest residentReq
...
+        /// <response code="500">There was a problem updating the records</response>
+        ///
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [HttpPatch]
+        [Route("residents")]
+        public IActionResult UpdatePerson([FromBody] UpdatePersonRequest request)
+        {
+            var response = _personUseCase.ExecutePatch(request);
+
+            if (response?.Message == "PersonNotFound")
```
JohnFarrellDev, line 143: Instead of this response being returned do we want to just throw an UpdatePersonException in the gateway if no person is found, then utilise try catch and return NotFound with the exception message.

This way UseCase doesn't have to return anything and we're aren't validating based on strings being specific values.

JohnFarrellDev (changes requested):

SocialCareCaseViewerApi.Tests/V1/Boundary/Request/GetResidentRequestTests.cs
```diff
@@ -0,0 +1,64 @@
...
+using FluentAssertions;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using System.Linq;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Boundary.Request
+{
+    [TestFixture]
+    public class GetResidentRequestTests
```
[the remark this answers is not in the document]
LBHTKarki: renamed

SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
```diff
@@ -121,6 +123,59 @@ public void AddNewResidentReturns500WhenAddressCouldNotBeInserted()
             response.Value.Should().Be("Address could not be inserted");
         }
 
+        [Test]
+        public void GetPersonByIdReturns200WhenSuccessfull()
+        {
+            GetPersonRequest request = new GetPersonRequest();
+            GetPersonResponse getPersonResponse = _fixture.Create<GetPersonResponse>();
```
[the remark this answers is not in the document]
LBHTKarki: removed

SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
```diff
@@ -121,6 +123,59 @@ public void AddNewResidentReturns500WhenAddressCouldNotBeInserted()
...
+
+            _mockPersonUseCase.Setup(x => x.ExecuteGet(It.IsAny<GetPersonRequest>())).Returns(response);
+
+            var result = _classUnderTest.GetPerson(request) as NotFoundResult;
+
+            result.StatusCode.Should().Be(404);
+        }
+
+        [Test]
+        public void UpdatePersonReturns200WhenSuccessfull()
```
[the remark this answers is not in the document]
LBHTKarki: updated

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
LBHTKarki: Implemented the exception approach to tidy this up.

SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs
LBHTKarki: I'll implement these changes in another PR as discussed on Slack

JohnFarrellDev (approved):
Looks good to me.

PR #210, opened 2021-04-20
Refactor get all warning notes endpoint path

Renny4Real:
## Link to JIRA ticket
https://hackney.atlassian.net/browse/SCS-717

We identified that the GET endpoint for retrieving all the warning notes created for a person should also follow a more RESTful pattern. This endpoint should return all warning notes for a specific person and should NOT imply that a warning note's ID is required.

This changes the path for retrieving all warning notes for a specific person to be:
`/residents/{personId}/warningNotes`

This also adds a change from a [PR](https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/206) that was merged into `master` to `development` (See commit dd5edb9)

#### _Checklist_

- [ ] Added comments to the README or updated, relevant documentation _(add a link to documentation)_, where necessary.

Test in staging that updated path works and created warning notes can be retrieved.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -563,23 +563,24 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         [Test]
         public void GetWarningNotesReturnsTheExpectedWarningNote()
         {
-            WarningNote warningNote = new WarningNote()
+            var testPersonId = _faker.Random.Int();
+            var differentPersonId = _faker.Random.Int();
+
+            var warningNote = new WarningNote
             {
-                PersonId = 12345
+                PersonId = testPersonId
             };
-            WarningNote wrongWarningNote = new WarningNote()
+
+            var wrongWarningNote = new WarningNote
             {
-                PersonId = 67890
+                PersonId = differentPersonId
             };
+
             DatabaseContext.WarningNotes.Add(warningNote);
             DatabaseContext.WarningNotes.Add(wrongWarningNote);
             DatabaseContext.SaveChanges();
 
-            var request = _fixture.Build<GetWarningNoteRequest>()
-                            .With(x => x.PersonId, warningNote.PersonId)
-                            .Create();
-
-            var response = _classUnderTest.GetWarningNotes(request);
+            var response = _classUnderTest.GetWarningNotes(testPersonId);
 
             response.Should().ContainSingle();
             response.Should().ContainEquivalentOf(warningNote);
@@ -589,22 +590,20 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         [Test]
         public void GetWarningNotesReturnsAnExceptionIfTheWarningNoteDoesNotExist()
         {
-            WarningNote warningNote = new WarningNote()
+            var testPersonId = _faker.Random.Int();
+            var differentPersonId = _faker.Random.Int();
+
+            var warningNote = new WarningNote
             {
-                PersonId = 12345
+                PersonId = testPersonId
             };
             DatabaseContext.WarningNotes.Add(warningNote);
             DatabaseContext.SaveChanges();
 
-            var request = new GetWarningNoteRequest()
-            {
-                PersonId = 67890
-            };
-
-            Action act = () => _classUnderTest.GetWarningNotes(request);
+            Action act = () => _classUnderTest.GetWarningNotes(differentPersonId);
 
             act.Should().Throw<DocumentNotFoundException>()
-                .WithMessage($"No warning notes found relating to person id {request.PersonId}");
+                .WithMessage($"No warning notes found relating to person id {differentPersonId}");
         }
         #endregion
     }
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -588,6 +588,42 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         }
 
         [Test]
+        public void GetWarningNotesReturnsAListOfWarningNotesForASpecificPerson()
+        {
+            var testPersonId = _faker.Random.Int();
+            var differentPersonId = _faker.Random.Int();
+
+            var firstNote = new WarningNote
+            {
+                PersonId = testPersonId,
+                Notes = "I am one note"
+            };
+
+            var secondNote = new WarningNote
+            {
+                PersonId = testPersonId,
+                Notes = "I am another note"
+            };
+
+            var separateWarningNote = new WarningNote
+            {
+                PersonId = differentPersonId
+            };
+
+            DatabaseContext.WarningNotes.Add(firstNote);
+            DatabaseContext.WarningNotes.Add(secondNote);
+            DatabaseContext.WarningNotes.Add(separateWarningNote);
+            DatabaseContext.SaveChanges();
+
+            var response = _classUnderTest.GetWarningNotes(testPersonId);
+
+            response.Count().Should().Be(2);
+            response.Should().ContainEquivalentOf(firstNote);
+            response.Should().ContainEquivalentOf(secondNote);
+            response.Should().NotContain(separateWarningNote);
+        }
+
+        [Test]
         public void GetWarningNotesReturnsAnExceptionIfTheWarningNoteDoesNotExist()
         {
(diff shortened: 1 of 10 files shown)


SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -388,19 +388,18 @@ public IActionResult PostWarningNote([FromBody] PostWarningNoteRequest request)
...
 
         /// <summary>
-        /// Get warning notes by person id
+        /// Get all warning notes created for a specific person
         /// </summary>
-        /// <param name="request"></param>
         /// <response code="200">Success. Returns warning notes related to the specified ID</response>
         /// <response code="404">No warning notes found for the specified ID</response>
-        [ProducesResponseType(typeof(List<WarningNote>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(ListWarningNotesResponse), StatusCodes.Status200OK)]
```
JohnFarrellDev, line 395: Is there a reason why we want to return `ListWarningNotesResponse` instead of `List<WarningNote>`

Considering `ListWarningNotesResponse` is a class with a single property of type `List<WarningNote>` the class does not seem required.

The different for the frontend is the json with ListWarningNotesResponse would look something like {warningNotes: [{note1}, {note2}]} .

With a List<WarningNote> the response would be more like [{note1}, {note2}].

I've noticed this pattern with a single property where the class does not need to exist elsewhere and in some places has been updated to delete the unrequired class from the codebase.

SocialCareCaseViewerApi/V1/UseCase/Interfaces/IWarningNoteUseCase.cs
```diff
@@ -1,13 +1,11 @@
...
 using SocialCareCaseViewerApi.V1.Boundary.Response;
-using SocialCareCaseViewerApi.V1.Domain;
 
 namespace SocialCareCaseViewerApi.V1.UseCase.Interfaces
 {
     public interface IWarningNoteUseCase
     {
         PostWarningNoteResponse ExecutePost(PostWarningNoteRequest request);
-        List<WarningNote> ExecuteGet(GetWarningNoteRequest request);
+        ListWarningNotesResponse ExecuteGet(long personId);
```
JohnFarrellDev, line 9: If you agree with SocialCareCareCaseViewerApiController comment this could be reverted.

SocialCareCaseViewerApi/V1/UseCase/WarningNoteUseCase.cs
```diff
@@ -22,12 +20,14 @@ public PostWarningNoteResponse ExecutePost(PostWarningNoteRequest request)
             return _databaseGateway.PostWarningNote(request);
         }
 
-        public List<WarningNote> ExecuteGet(GetWarningNoteRequest request)
+        public ListWarningNotesResponse ExecuteGet(long personId)
```
JohnFarrellDev, line 23: If you agree with SocialCareCareCaseViewerApiController comment this could be reverted.

SocialCareCaseViewerApi/V1/Boundary/Response/ListWarningNotesResponse.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
```
JohnFarrellDev, line 1: Maybe this class can be deleted per the SocialCareCareCaseViewerApiController comment.

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -563,48 +563,83 @@ private Team SaveTeamToDatabase(Team team)
...
             response.Should().ContainEquivalentOf(warningNote);
             response.Should().NotContain(wrongWarningNote);
         }
 
         [Test]
-        public void GetWarningNotesReturnsAnExceptionIfTheWarningNoteDoesNotExist()
+        public void GetWarningNotesReturnsAListOfWarningNotesForASpecificPerson()
         {
-            WarningNote warningNote = new WarningNote()
+            var testPersonId = _faker.Random.Int();
```
JohnFarrellDev, line 593: This could be tider if we used TestHelpers.CreatePerson() and maybe created a function in TestHelpers for Creating Warning Notes (that can take a personId as a potential paramter)

Could also use Helper functions in test above.

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -388,19 +388,18 @@ public IActionResult PostWarningNote([FromBody] PostWarningNoteRequest request)
...
         [HttpGet]
-        [Route("warningnotes/{id}")]
-        public IActionResult GetWarningNote([FromQuery] GetWarningNoteRequest request)
+        [Route("residents/{personId}/warningNotes")]
+        public IActionResult ListWarningNotes(long personId)
```
JohnFarrellDev, line 398: To me `GetWarningNotes` makes more sense, the function name describes what it is doing, List is referring to the type being returned and not the functionality?

You might disagree though which is ok.

SocialCareCaseViewerApi/V1/Infrastructure/WarningNote.cs
```diff
@@ -16,46 +16,49 @@ public class WarningNote : IAuditEntity
...
-        [Required]
         public DateTime? ReviewDate { get; set; }
 
         [Column("end_date")]
         public DateTime? EndDate { get; set; }
 
         [Column("individual_notified")]
-        [Required]
-        public Boolean DisclosedWithIndividual { get; set; }
+        public bool DisclosedWithIndividual { get; set; }
```
JohnFarrellDev, line 28: I know a lot of this was made before you touched this code but it could be really good to do define fields that are not required as nullable, this should help us avoid NullReferenceExceptions in the code.

I am happy for this to be part of a separate PR/Refactor piece as it does feel outside scope of this work, I just wanted to make a note of it here.

JohnFarrellDev (changes requested):
The new changes to be more RESTful is really good as is the exception handling.

I think some useful changes from before maybe got reverted by accident in this PR which is what I think most of the comments are referring to.

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
Renny4Real: Not sure what the best practice is, but I've seen a lot of APIs return a list in the format of { key: ["item", "item2"] }.
According to our [swagger documentation](https://dr03nduqxh.execute-api.eu-west-2.amazonaws.com/staging/swagger/index.html), there are some endpoints returning responses in this format and some that don't.

I don't think it's a bad format to follow (it looks tidier, in my opinion 😅 )

So maybe this could be something we could discuss as a team? Since we have API documentation containing GET endpoints that don't follow a strict format.

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
Renny4Real: I agree 👍🏾 . I think we can do that in a separate PR.

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
Renny4Real: I can see both 🤔 
I was thinking from the perspective of `List`as a verb. So this lists all the warning notes created for a person, and maybe there would be a separate action to get a single warning note, but I don't have any strong opinions on this aside from whatever would clearly convey the idea behind the code.

JohnFarrellDev: I think this is fine.

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
JohnFarrellDev: The reason I don't like is that the List class already exists that behaves generically with any object type we give it.

It feels "wrong" to make a class called ListX when List<X> is just as easy but doesn't require the creation of a wrapper class.

I agree a quick team discussion on this would be useful so the backend can be consistent.

SocialCareCaseViewerApi/V1/Infrastructure/WarningNote.cs
Renny4Real: Happy for this to be a separate PR for refactoring 👍🏾

JohnFarrellDev (approved):
Good to go :)

PR #215, opened 2021-04-22
Rename fix historic case notes response feature flag

chubberlisk:

Due to a mismatch of Terraform in `master` and `development` and
CircleCI running Terraform for staging on both branches, Terraform
is unable to create (again) the feature flag for fixing the historic
case note response. As a result, we need to rename it as it was
previously created but then scheduled for deletion because `master`
has no Terraform related to it and you can't create a new secret with
the same name that's scheduled for deletion.

image

This PR renames the feature flag for fixing historic case notes response and removes the show historic data feature flag.


JohnFarrellDev (approved):
looks good

PR #219, opened 2021-04-22
Add GET endpoint to get a specific warning note with its associated reviews

Renny4Real:
**Any feedback/suggestions would be appreciated 🙂** 

## Link to JIRA ticket

[Create GET endpoint for retrieving a specific warning note with its associated reviews
](https://hackney.atlassian.net/browse/SCS-723)

## Describe this PR

When a caseworker views a specific warning note as part of the reviewing process, they should see a list of previous reviews for that specific warning note.

Created an endpoint `warningNote/{warningNoteId}` which, when called, would return the warning note with the specified ID and a list of all reviews (if any) saved for that warning note.

#### _Checklist_

- [ ] Added comments to the README or updated relevant documentation _(add a link to documentation)_, where necessary.

Test calling the endpoint in Staging returns data as desired.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
@@ -374,6 +374,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
                 DisclosedWithIndividual = dbWarningNote.DisclosedWithIndividual,
                 DisclosedDetails = dbWarningNote.DisclosedDetails,
                 Notes = dbWarningNote.Notes,
+                ReviewDate = dbWarningNote.ReviewDate?.ToString("s"),
                 NextReviewDate = dbWarningNote.NextReviewDate?.ToString("s"),
                 NoteType = dbWarningNote.NoteType,
                 Status = dbWarningNote.Status,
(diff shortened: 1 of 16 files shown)


SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
```diff
@@ -342,5 +342,54 @@ public void CanMapPersonDetailsToGetPersonResponse()
 
             result.Should().BeEquivalentTo(expectedResponse);
         }
+
+        [Test]
+        public void MapWarningNoteToResponseReturnsAnObjectWithFormattedDates()
```
JohnFarrellDev, line 347: Could we have an additional test that also makes sure null is handled correctly for any property that is nullable.

We probably want to do this consistently going forward to catch those annoying NullReferenceExceptions.

JohnFarrellDev (approved):
This was fantastic, code is really clean and readable :+1: 

Only 1 minor comment and I think it's something we all need to be doing as a team.

SocialCareCaseViewerApi.Tests/V1/Factories/ResponseFactoryTests.cs
Renny4Real: Thanks 🙌🏾 . 
I'll need to merge another PR (#230) before this one to try and fix a bug in Staging. 😅 
I'll try adding a test for checking this after rebasing.

PR #222, opened 2021-04-23
Fix sorting of newly created case notes in records history

chubberlisk:

For newly created case notes, they appear at the bottom of the records history when they should appear at the top. This is because date of event was missing from the request which meant there was no date to sort by.

Also when testing this in staging, we also discovered a smol bug with the date of birth causing an issue when creating a case note with a null DOB.

This PR:

- fixes the null date of birth by doing a null check when doing `To.String()`
- makes sure we use date of event when creating a case note
- allows sorting for ISO format date of event

#### _Checklist_

Check that these changes fix the issues in staging and add tests for `ToEntity` for `CreateCaseNoteRequest`.

## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCS-592

### DIFF
--- a/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
+++ b/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
@@ -175,7 +175,7 @@ namespace SocialCareCaseViewerApi.V1.Factories
             GenericCaseNote note = new GenericCaseNote()
             {
                 Timestamp = DateTime.Now.ToString("dd/MM/yyy hh:mm"),
-                DateOfBirth = request.DateOfBirth.ToString(),
+                DateOfBirth = request.DateOfBirth?.ToString(),
                 FirstName = request.FirstName,
                 LastName = request.LastName,
                 FormName = request.FormName,
--- a/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
+++ b/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
@@ -175,7 +175,7 @@ namespace SocialCareCaseViewerApi.V1.Factories
             GenericCaseNote note = new GenericCaseNote()
             {
                 Timestamp = DateTime.Now.ToString("dd/MM/yyy hh:mm"),
-                DateOfBirth = request.DateOfBirth?.ToString(),
+                DateOfBirth = request.DateOfBirth?.ToString("dd/MM/yyy"),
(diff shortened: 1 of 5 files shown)


SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -175,7 +175,8 @@ public static CaseNotesDocument ToEntity(this CreateCaseNoteRequest request)
             GenericCaseNote note = new GenericCaseNote()
             {
                 Timestamp = DateTime.Now.ToString("dd/MM/yyy hh:mm"),
-                DateOfBirth = request.DateOfBirth.ToString(),
+                DateOfBirth = request.DateOfBirth?.ToString(),
```
Renny4Real, line 178: Do we want to return the Datetime properties here as ISO format?

Renny4Real (approved):
👍🏾 Looks good

SocialCareCaseViewerApi/V1/Gateways/ProcessDataGateway.cs
```diff
@@ -240,19 +240,21 @@ public IOrderedEnumerable<CareCaseData> SortData(string sortBy, string orderBy,
...
+                    if (success) return timeStamp;
                 }
                 else
                 {
                     bool success = DateTime.TryParseExact(x.DateOfEvent, "dd/MM/yyyy", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime dateOfEvent);
-                    if (success) dt = dateOfEvent;
+                    if (success) return dateOfEvent;
+
+                    bool successForISOFormat = DateTime.TryParseExact(x.DateOfEvent, "O", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime dateOfEventISOFormat);
+                    if (success) return dateOfEventISOFormat;
```
JohnFarrellDev, line 254: Should this be `if (successForISOFormat)`

LBHTKarki: well spotted, thank you! fixed

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
LBHTKarki: We have aligned this with existing data i.e. used dd/mm/yyyy format

JohnFarrellDev (approved):
Looks good

Renny4Real (approved):

PR #224, opened 2021-04-23
align the response to id and not personid for POST resident

JohnFarrellDev:
## Link to JIRA ticket

No ticket.

We want more consistency in properties returned by backend.
In this case return Id when creating a resident instead of PersonId.

#### _Checklist_

Deploy to prod if it looks good on staging.


sirLisko (approved):
Looks good 🚀

PR #227, opened 2021-04-23
add support for ISO date format in sorting

LBHTKarki:
## Describe this PR

Add ISO date format yyyy-mm-dd support to case search sorting. This will support the date format used by the front end for new case notes.


chubberlisk (approved):
:shipit:

PR #228, opened 2021-04-26
Update/restricted flag to add resident

LBHTKarki:
## Link to JIRA ticket
This is part of the endpoints alignment for create/edit/view person feature

## Describe this PR

Add restricted property to AddNewResidentRequest object. This will enable front end to use the same form for create and edit person pages.

Added a new property to the request. It's not mandatory and will be set to "N" by default, so this can be deployed whether the front end provides the value or not.

I've also removed some commented out code and obsolete boundary object property tests. 

#### _Checklist_

- [x ] Added tests to cover all new production code

Inform front end devs that the restricted value can now be provided when creating new person records.

### DIFF
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/AddNewResidentRequest.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/AddNewResidentRequest.cs
@@ -61,5 +61,9 @@ namespace SocialCareCaseViewerApi.V1.Boundary.Requests
         [Required]
         [EmailAddress]
         public string CreatedBy { get; set; }
+
+        [MaxLength(1)]
+        [RegularExpression("(?i:^Y|N)", ErrorMessage = "The restricted must be 'Y' or 'N' only.")]
+        public string Restricted { get; set; } = "N";
     }
 }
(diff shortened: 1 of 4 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/AddNewResidentRequest.cs
```diff
@@ -61,5 +61,9 @@ public string EmailAddress
         [Required]
         [EmailAddress]
         public string CreatedBy { get; set; }
+
+        [MaxLength(1)]
+        [RegularExpression("(?i:^Y|N)", ErrorMessage = "The restricted must be 'Y' or 'N' only.")]
+        public string Restricted { get; set; } = "N";
```
chubberlisk, line 67: How does this validation work? Do we have to catch an exception in the controller for the endpoint to return a 400 bad request? Just wondering if we have to do something similar like creating allocations:

https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/f2e3e4943e153f402ad890f9470b722b346d123d/SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs#L229-L237

chubberlisk (commented):
Generally looks good to me, just have a question around how the validation works.

SocialCareCaseViewerApi/V1/Boundary/Requests/AddNewResidentRequest.cs
LBHTKarki: This is using out of the box .NET validation attributes, so the validation and error messages are handled automatically by the framework based on the model. Allocation is using Fluent validation library now which requires some additional setup. We are slowly refactoring stuff to use that one, but the person and most other end points are still using the default approach.

chubberlisk: Ahh cooool, that makes sense. Thanks for the explanation!

chubberlisk (approved):
:shipit:

PR #231, opened 2021-04-28
use utc date format in dates created by faker

LBHTKarki:
This update resolves an issue where some tests were failing on Windows and Mac.

Some tests using Date method from Faker library were randomly failing on Windows and Mac.  

Update the test setup so dates generated by Faker library use UTC format. 

Ensure all tests involving dates generated by Faker library are using UTC format


chubberlisk (approved):
🛠️

PR #232, opened 2021-04-28
Removes status from the POST request body for creating a warning note

Renny4Real:
When we first create a warning note, we want to set the status to "open".

Since we set this value upon the creation of a new warning note,
this removes the value as a request parameter for POST.


JohnFarrellDev (approved):
Makes sense and looks good to me

PR #236, opened 2021-05-06
Fix get workers

JohnFarrellDev:
## Link to JIRA ticket

No ticket.

Updating Worker class has broken the ability to get a worker, this is intended as a fix.

Remove requirement for IsActive property and make LastModifiedBy nullable.

#### _Checklist_


SocialCareCaseViewerApi/V1/Infrastructure/Worker.cs
```diff
@@ -30,7 +30,6 @@ public class Worker : IAuditEntity
         public string LastName { get; set; } = null!;
 
         [Column("is_active")]
-        [Required]
         public bool IsActive { get; set; }
```
LBHTKarki, line 33: Should we add an update to the schema file to update all null is_active values to true i.e. 
`update dbo.sccv_worker set is_active = true where is_active isnull;`

LBHTKarki (approved):
Works fine as long as the database has been handled in advance.

SocialCareCaseViewerApi/V1/Infrastructure/Worker.cs
JohnFarrellDev: Implemented

LBHTKarki (approved):

PR #242, opened 2021-05-07
Return an empty list if no warning notes for a person is found

Renny4Real:
## Describe this PR

From testing the feature on staging, if a person has no warning notes, an `Oops, an error occurred` message is displayed. 
From checking the response from the backend, we found that if a person has no warning notes, the response was returning `warningNotes: null` instead of `warningNotes: [ ]`.

If a person has no warning notes, the gateway would return an empty list rather than null.

#### _Checklist_

- [ ] Added comments to the README or updated relevant documentation _(add a link to documentation)_, where necessary.

Test that the above error message does not pop up if a person has no warning notes.


chubberlisk (approved):
:shipit:

PR #250, opened 2021-05-12
use 24h format in timestamp

LBHTKarki:
## Link to JIRA ticket

This is related to the sorting issue addresses in https://hackney.atlassian.net/browse/SCT-8

Currently the timestamp is not using 24hr format so the records end up with times that are causing issues with sorting

Change the timestamp to use UTC time and 24h hour format. Further development work is required to standardise all date times to ISO, but this fix will improve the captured data in the meantime.

### DIFF
--- a/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
+++ b/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
@@ -180,7 +180,7 @@ namespace SocialCareCaseViewerApi.V1.Factories
         {
             GenericCaseNote note = new GenericCaseNote()
             {
-                Timestamp = DateTime.UtcNow.ToString("dd/MM/yyyy H:mm:ss"),
+                Timestamp = DateTime.UtcNow.ToString("dd/MM/yyyy HH:mm:ss"),
                 DateOfBirth = request.DateOfBirth?.ToString("dd/MM/yyy"),
                 DateOfEvent = request.DateOfEvent?.ToString(),
                 FirstName = request.FirstName,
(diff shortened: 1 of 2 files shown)


SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -180,7 +180,7 @@ public static CaseNotesDocument ToEntity(this CreateCaseNoteRequest request)
         {
             GenericCaseNote note = new GenericCaseNote()
             {
-                Timestamp = DateTime.Now.ToString("dd/MM/yyy hh:mm"),
+                Timestamp = DateTime.UtcNow.ToString("dd/MM/yyyy H:mm:ss"),
```
chrishutchinson, line 183: Should a test also need to change to reflect this?

LBHTKarki: This is one of the tech debt areas because we don't have any test coverage for this. I'll add some now so it's not missing completely

LBHTKarki: @chrishutchinson I've added two tests that cover the object conversion as well as the date time format for timestamp. It's not perfect since checking the date time after it's been converted to string is bit of a challenge in terms of checking whether it was created using 24h format or not. These tests should protect against the main format changes at least.

chrishutchinson: Nice one @LBHTKarki! That makes, and looks good, approved ✅

chrishutchinson (approved):

PR #253, opened 2021-05-13
Endpoint to get teams

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-117)

We want to be able get a list of teams with the team context (already existed), list of size 1 for getting team by team id, list of size 1 for getting team by team name.

Updated the Query received by GET /api/v1/teams to also accept name or id. All fields are optional but one must be provided. If multiple are provided the search priority highest to lowest is Id, Name, Context.

Updated Teams usecase for ExecuteGet to now utilise the requested team name or id if provided.

Created gateway method for getting team by team name.
Updated existing gateway method to get team by team id, used to return a list<Team> but now returns a single Team.

Updated ExecutePost to throw an exception if trying to create a team with a name that is is already taken.

#### _Checklist_

We should look to deploy with the endpoint for creating teams.


chubberlisk (approved):
Looks good to me!

PR #260, opened 2021-05-18
CaseNote now added when warning note updated

LBHBReynolds-Carr:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-136

Reviewing or ending of a warning note does not appear in the record history at all. 

Case notes were not being added as part of the PatchWarningNotes method, the method has been updated to create and insert CaseNotes into the MongoDB database.

#### _Checklist_

- [x ] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature
- [x ] Added tests to cover all new production code
- [ x] Added comments to the README or updated relevant documentation _(add link to documentation)_, where necessary.
- [x ] Checked all code for possible refactoring
- [x ] Code pipeline builds correctly

N/A

### DIFF
--- a/SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
+++ b/SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
@@ -825,6 +825,28 @@ namespace SocialCareCaseViewerApi.V1.Gateways
             var review = PostWarningNoteReview(request);
             _databaseContext.WarningNoteReview.Add(review);
             _databaseContext.SaveChanges();
+
+            var dt = DateTime.Now;
+
+            var note = new WarningNoteCaseNote
+            {
+                FirstName = person.FirstName,
+                LastName = person.LastName,
+                MosaicId = person.Id.ToString(),
+                Timestamp = dt.ToString("dd/MM/yyyy H:mm:ss"),
+                Note = $"{dt.ToShortDateString()} | Warning Note | {((request.Status == "closed") ? "Warning note against this person ended" : "Warning note against this person reviewed")}",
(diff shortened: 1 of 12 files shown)


SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -825,6 +825,28 @@ public void PatchWarningNote(PatchWarningNoteRequest request)
...
+                WarningNoteId = warningNote.Id.ToString(),
+                WorkerEmail = request.ReviewedBy
+            };
+
+            var caseNotesDocument = new CaseNotesDocument
+            {
+                CaseFormData = JsonConvert.SerializeObject(note)
+            };
+
+            _ = _processDataGateway.InsertCaseNoteDocument(caseNotesDocument).Result;
```
Renny4Real, line 849: Why do we assign a `discard` variable here since this method is a `void` return? 
Should we be returning this?

Renny4Real (commented):
Looks good—just a quick question.

Many of these look to be linting changes unrelated to the patch warning note, wondering if we need to add anything to our `.editorconfig` or something around linting consistency between editors? 🤔

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
LBHBReynolds-Carr: The method `InsertCaseNoteDocument` returns `Task<string>` and not `void`, so we need to assign the result of the method to a variable. As we don't need the variable, it's assigned to a `discard`.

PR #267, opened 2021-05-24
refactor(SCT-260): remove tests do not provide value

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-260)

We want all our tests to provide value.
This PR removes a lot of tests that are seen to not provide value. 
It also tidies up some tests that are kept.

#### _Checklist_

Look to deploy because why not


SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
```diff
@@ -66,8 +66,8 @@ public void ListContactsReturns200WhenSuccessful()
             var response = _classUnderTest.ListContacts(residentQueryParam, 2, 3) as OkObjectResult;
 
             response.Should().NotBeNull();
-            response.StatusCode.Should().Be(200);
-            response.Value.Should().BeEquivalentTo(residentInformationList);
+            response?.StatusCode.Should().Be(200);
```
LBHTKarki, line 69: Is it preferable to use this pattern i.e. check that response is not null rather than let the test throw an exception? A bit tidier I suppose. Also there's a null check above for the response itself, so perhaps the second check is not necessary.

JohnFarrellDev: I think it's tidier this way and the test will fail if the response is null. (I know it would also fail the NotBeNull check first)

If the response is null then it would just be `null.Should().Be(200)` which will fail.

This is the neatest way that also keeps the compiler happy with null reference warnings.

LBHTKarki (approved):
Good clean up work 👍

PR #270, opened 2021-05-25
Fix/SCT 261 allow worker no team

Miles-Alford:
## Link to JIRA ticket

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=59&modal=detail&selectedIssue=SCT-261

The back-end requires a worker to have a team and a role however some pre-existing workers may not have these attributes, if they do not have a team they are unable to be edited. 

We have removed the validation which requires a worker to have a role and also have at least one team within the `Teams` array.

#### _Checklist_

There is a bug which stops the `Teams` list from being empty, this will be resolved in a future PR


JohnFarrellDev (approved):
:+1:

PR #271, opened 2021-05-27
feature(SCT-342): Create common mongodb gateway

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-342)

We need a way to do our standard operations with Mongo for any object type.

Created a generic MongoGateway class that acts as a thin layer around Mongo to provide us with standard CRUD functionality.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using FluentAssertions;
 using MongoDB.Bson.Serialization.Attributes;
 using NUnit.Framework;
@@ -154,8 +153,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
             {
                 retrievedObject =
                     _mongoGateway.LoadMultipleRecordsByProperty<TestObjectForMongo, string>(
-                        "test-collection-name", "Property1", _testObjectForMongo.Property1)
-                        .ToList();
+                        "test-collection-name", "Property1", _testObjectForMongo.Property1);
             }
             retrievedObject?.Count.Should().Be(2);
         }
@@ -169,8 +167,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
             {
                 retrievedObject =
                     _mongoGateway.LoadMultipleRecordsByProperty<TestObjectForMongo, string>(
-                            "test-collection-name", "Property1", _testObjectForMongo.Property1)
-                        .ToList();
+                            "test-collection-name", "Property1", _testObjectForMongo.Property1);
             }
             retrievedObject?.Should().BeEmpty();
         }
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
@@ -1,0 +1,240 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using MongoDB.Bson.Serialization.Attributes;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.V1.Gateways;
+
+#nullable enable
+namespace SocialCareCaseViewerApi.Tests.V1.Gateways
+{
+    public class TestObjectForMongo
+    {
+        [BsonId]
+        public Guid Id { get; set; }
+        public string Property1 { get; set; } = null!;
+        public List<TestObjectForMongo>? Property2 { get; set; }
+        public TestObjectForMongo? Property3 { get; set; }
+    }
+
+    [TestFixture]
+    public class MongoGatewayTests
+    {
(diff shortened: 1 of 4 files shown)


SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        public void LoadMultipleRecordsByCustomerPropertyWithNoMatchReturnsEmptyList()
+        {
+            List<TestObjectForMongo>? retrievedObject = null;
+
+            if (_testObjectForMongo != null)
+            {
+                retrievedObject =
+                    _mongoGateway.LoadMultipleRecordsByProperty<TestObjectForMongo, string>(
+                            "test-collection-name", "Property1", _testObjectForMongo.Property1)
+                        .ToList();
```
chubberlisk, line 173: Do we need to do `ToList()` here? Looks like the `_mongoGateway.LoadMultipleRecordsByProperty` does it already.

chubberlisk (commented):
This is really cool - TIL about generics in C#, thank you! 💯 

In the tests, I noticed we use `_mongoGateway.InsertRecord` to seed the MongoDB. This to me means we're testing two methods of the MongoGateway in a test e.g. in `CanLoadMultipleRecordsByACustomProperty`, we're testing `InsertRecord` and `LoadMultipleRecordsByProperty` because we're using both in a test. Therefore, if there was something wrong with `InsertRecord`, the `LoadMultipleRecordsByProperty` tests would failed as well as the `InsertRecord` ones which makes things a little more difficult to debug. I'm wondering whether it's worth refactoring and having a similar setup as our PostgreSQL database tests, where we directly insert and teardown at the database-level to ensure our tests are isolated. It might not be worth the effort but worth thinking about. What do you think?

SocialCareCaseViewerApi.Tests/V1/Gateways/MongoGatewayTests.cs
JohnFarrellDev: Good suggestion, I had to update the interface to say a List<> was being returned and not an IEnumerable<>.

I'm all about performance gains.

JohnFarrellDev:
> This is really cool - TIL about generics in C#, thank you! 
> 
> In the tests, I noticed we use `_mongoGateway.InsertRecord` to seed the MongoDB. This to me means we're testing two methods of the MongoGateway in a test e.g. in `CanLoadMultipleRecordsByACustomProperty`, we're testing `InsertRecord` and `LoadMultipleRecordsByProperty` because we're using both in a test. Therefore, if there was something wrong with `InsertRecord`, the `LoadMultipleRecordsByProperty` tests would failed as well as the `InsertRecord` ones which makes things a little more difficult to debug. I'm wondering whether it's worth refactoring and having a similar setup as our PostgreSQL database tests, where we directly insert and teardown at the database-level to ensure our tests are isolated. It might not be worth the effort but worth thinking about. What do you think?

I'm happy you find it cool :)

You're right that Inserting failing would cause multiple tests to fail which would make debugging slightly harder. I do think people would still figure it out pretty quickly. 

Is the alternative to update the DatabaseTests.cs file so that it is creating collection, inserting record, deleting collection using the [SetUp] and [TearDown] functions. If for some reason the inserting failed the test such as CanLoadMultipleRecordsByACustomProperty would also not work for a unclear reason.

My feeling is it's probably fine as it is and making changes only moves the same problem somewhere else.

SocialCareCaseViewerApi/V1/Gateways/MongoGateway.cs
```diff
@@ -0,0 +1,76 @@
...
+        public List<T> LoadRecords<T>(string collectionName)
+        {
+            var collection = _mongoDatabase.GetCollection<T>(collectionName);
+            return collection.Find(new BsonDocument()).ToList();
+        }
+
+        public T LoadRecordById<T>(string collectionName, Guid id)
+        {
+            var collection = _mongoDatabase.GetCollection<T>(collectionName);
+            var filter = Builders<T>.Filter.Eq("Id", id);
```
chubberlisk, line 58: What's the difference between `"Id"` and `"_id"`? 🤔 In `DeleteRecordById`, we filter by `_id` instead.

chubberlisk (approved):
I'm happy to go with what you reckon! 👍

SocialCareCaseViewerApi/V1/Gateways/MongoGateway.cs
JohnFarrellDev: I'm honestly surprised it was working, the Mongo ORM must be taking the field "Id" and converting it to "_id". 
I'll update it to use "_id" though.

The way Mongo works with C# is that you give a property on a class the [BsonId] directive, this tells Mongo it is the Id for the object and maps it to the "_id" field in mongo.

I should later make a little change to the ReadMe for debugging Mongo, you can point MongoDb Compass to our docker mongo instance and see the objects being created and removed in our tests. I'm sure at some point it will help someone to debug something if they can see what is actually being added to Mongo. (this is really just a reminded for me :p )

chubberlisk: That sounds useful especially because I have no idea what MongoDB Compass is! 😂

PR #273, opened 2021-05-28
fix(SCT-381): get 500 error searching for some workers

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-381)

Trying to get the details of certain workers is failing, this is probably due to poor null handling in the case of WorkerTeams or Allocations.

We have introduced better null handling checks and tested the conversion of a Infrastructure Worker object to a Domain Worker object when the teams or allocations are null.

The changes that fix the bug are in the EntityFactory.cs where we now handle null better. 
Change to Worker.cs is to signal that WorkerTeams and Allocations may be null.
Other changes are to tidy up our tests.

#### _Checklist_

Confirm the changes don't also require us to update the schema.sql file and do a manual update of the database too!
We just need to confirm this in staging environment before deploying anything, if we can create and get a worker it is fine basically.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Collections.Generic;
-using System.Dynamic;
 using System.Globalization;
+using System.Linq;
 using AutoFixture;
 using Bogus;
 using FluentAssertions;
@@ -42,44 +42,20 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         [Test]
         public void CanMapWorkerFromInfrastructureToDomainWithoutTeamDetails()
         {
-            var email = _faker.Internet.Email();
-            var firstName = _faker.Name.FirstName();
-            var lastName = _faker.Name.LastName();
-            var id = 1;
-            var role = _faker.Random.Word();
-            int allocationCount = 1;
-            var contextFlag = "A";
-            var createdBy = _faker.Internet.Email();
-            var dateStart = DateTime.Now;
-
-            var dbWorker = new dbWorker()
-            {
-                Email = email,
-                FirstName = firstName,
-                LastName = lastName,
-                Id = id,
-                Role = role,
-                ContextFlag = contextFlag,
-                CreatedBy = createdBy,
-                DateStart = dateStart,
-                Allocations = new List<AllocationSet>() {
-                    new AllocationSet() { Id = 1, PersonId = 2, CaseStatus = "Open" },
-                    new AllocationSet() { Id = 2, PersonId = 3, CaseStatus = "Closed" }
-                }
-            };
+            var dbWorker = TestHelpers.CreateWorker();
 
             var expectedResponse = new Worker()
             {
-                FirstName = firstName,
-                LastName = lastName,
-                Id = id,
-                AllocationCount = allocationCount,
-                Email = email,
-                Role = role,
-                ContextFlag = contextFlag,
-                CreatedBy = createdBy,
-                DateStart = dateStart,
-                Teams = null
+                FirstName = dbWorker.FirstName,
+                LastName = dbWorker.LastName,
+                Id = dbWorker.Id,
+                AllocationCount = dbWorker.Allocations?.Count(allocation => allocation.CaseStatus.ToUpper() == "OPEN") ?? 0,
+                Email = dbWorker.Email,
+                Role = dbWorker.Role,
+                ContextFlag = dbWorker.ContextFlag,
+                CreatedBy = dbWorker.CreatedBy,
+                DateStart = dbWorker.DateStart,
+                Teams = new List<Team>()
             };
 
             dbWorker.ToDomain(false).Should().BeEquivalentTo(expectedResponse);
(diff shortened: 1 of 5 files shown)


SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
```diff
@@ -88,56 +64,44 @@ public void CanMapWorkerFromInfrastructureToDomainWithoutTeamDetails()
...
+                    Role = workers[0].Role,
+                    ContextFlag = workers[0].ContextFlag,
+                    CreatedBy = workers[0].CreatedBy,
+                    FirstName = workers[0].FirstName,
+                    LastName = workers[0].LastName,
+                    DateStart = workers[0].DateStart,
+                    AllocationCount =
+                        workers[0].Allocations?.Count(allocation => allocation.CaseStatus.ToUpper() == "OPEN") ?? 0,
+                    Teams = workers[0].WorkerTeams?.Select(x =>
+                        new Team() {Id = x.Team.Id, Name = x.Team.Name, Context = x.Team.Context}).ToList() ?? new List<Team>()
```
chubberlisk, line 87: Possibly a silly question: Do we still need to do the null checks if the first worker will always have allocations and teams? Looks like the test helper will always create 3 of them.

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
```diff
@@ -143,9 +145,29 @@ public static (UpdateAllocationRequest, Worker, Worker, InfrastructurePerson, Te
...
+
+            return new Faker<AllocationSet>()
+                .RuleFor(a => a.Id, f => id ?? f.UniqueIndex + 1)
+                .RuleFor(a => a.PersonId, f => personId ?? f.UniqueIndex + 1)
+                .RuleFor(a => a.WorkerId, f => workerId ?? f.UniqueIndex + 1)
+                .RuleFor(a => a.TeamId, f => teamId ?? f.UniqueIndex + 1)
+                .RuleFor(a => a.AllocationStartDate, f => f.Date.Past())
+                .RuleFor(a => a.AllocationEndDate, f => f.Date.Past())
+                .RuleFor(a => a.CaseStatus, f => f.PickRandom(caseStatusChoices))
+                .RuleFor(a => a.CaseClosureDate, f => f.Date.Past());
```
chubberlisk, line 166: Can we add `.ToUniversalTime()` the date attributes? I think this caused tests to fail for Windows and MacOS before.

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
```diff
@@ -281,10 +303,13 @@ public static PersonOtherName CreatePersonOtherName(long? personId = null)
             int? workerId = null
         )
         {
+            var team = CreateTeam();
+
             return new Faker<WorkerTeam>()
                 .RuleFor(t => t.Id, f => f.UniqueIndex)
                 .RuleFor(t => t.WorkerId, f => workerId ?? f.UniqueIndex)
-                .RuleFor(t => t.TeamId, f => f.UniqueIndex);
+                .RuleFor(t => t.TeamId, f => f.UniqueIndex)
```
chubberlisk, line 311: QQ: When should we just use `f.UniqueIndex` or `f.UniqueIndex + 1`?

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -170,7 +170,7 @@ public static dbPhoneNumber ToEntity(this PhoneNumber number, long personId, str
         {
             return new dbPhoneNumber
             {
-                Number = number.Number.ToString(),
+                Number = number.Number,
```
chubberlisk, line 173: Why was this change made? 🤔

chubberlisk (approved):
Looks good to me! Just some smol questions and suggestions.

SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
JohnFarrellDev: You're right that at runtime we will always have 3 WorkerTeams so therefore workers[0].WorkerTeams will never be null.

However, the compiler does not know if WorkerTeams will exist as it can be null, therefore at a minimum this deals with a compiler warning.

If something can be null I think it is good to handle null even when we are confident then it won't actually be null. 

This is just test code but NullReferenceExceptions have been the cause of probably most of our annoying backend bugs so far.

This is why I really like the use of #nullable as we get to be explicit about what can and can't be null then are forced to handle null correctly.

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
JohnFarrellDev: It was the last warning I was getting in the file and I wanted that Rider green tick.

The type of number.Number is already a string (even if the naming of number.Number might suggest otherwise), calling .ToString() against a string just returns itself so it could be removed.

SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
chubberlisk: Ahhh okay, that makes sense! Still remembering we're in C# world. 😝

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
chubberlisk: Wicked, makes sense! ✅

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
JohnFarrellDev: UniqueIndex seems to start at 0 (and increments by 1 every time it is assigned), there are cases in our database where the minimum accepted value is 1 so the `f.UniqueIndex + 1` is to help some GatewayTests in which we actually add something to PostgreSQL.

It would probably be nice to have this consistently everywhere. Clearly it's also a little confusing so could be documented why.

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
JohnFarrellDev: Done

PR #276, opened 2021-06-02
feature(SCT-256): Post submission API

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-256)

The ability to create a form against a resident for a worker.

Created a new POST endpoint for the /submissions API, endpoint takes the workers email, resident Id and the id of the form.
From that request we then get the relevant Worker and Resident information from our database and populate a new empty but valid CaseSubmission into our MongoDb database.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/UseCase/SubmissionUseCaseTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/UseCase/SubmissionUseCaseTests.cs
@@ -17,6 +17,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.UseCase
         private Mock<IDatabaseGateway> _mockDatabaseGateway;
         private Mock<IMongoGateway> _mockMongoGateway;
         private SubmissionsUseCase _submissionsUseCase;
+        private const string CollectionName = "resident-case-submissions";
 
         [SetUp]
         public void SetUp()
@@ -25,7 +26,6 @@ namespace SocialCareCaseViewerApi.Tests.V1.UseCase
             _mockMongoGateway = new Mock<IMongoGateway>();
             _submissionsUseCase = new SubmissionsUseCase(_mockDatabaseGateway.Object, _mockMongoGateway.Object);
         }
-        // private const string CollectionName = "resident-case-submissions";
 
         [Test]
         public void ExecutePostSuccessfully()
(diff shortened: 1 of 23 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseSubmissionRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+
+#nullable enable
+{
+    public class CreateCaseSubmissionRequest
+    {
+        [JsonPropertyName("formId")]
+        public int FormId { get; set; }
+
+        [JsonPropertyName("socialCareId")]
+        public int ResidentId { get; set; }
```
chubberlisk, line 14: Are we using `ResidentId` because that's what we're currently using in MongoDB? Just wondering if we should name this `PersonId`.

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseSubmissionRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public class CreateCaseSubmissionRequestValidator : AbstractValidator<CreateCaseSubmissionRequest>
+    {
+        public CreateCaseSubmissionRequestValidator()
+        {
+            RuleFor(s => s.FormId)
+                .NotNull().WithMessage("Provide a form ID");
+            RuleFor(s => s.ResidentId)
+                .NotNull().WithMessage("Provided an ID for the resident");
```
chubberlisk, line 27:
```suggestion
                .NotNull().WithMessage("Provide an ID for the resident");
```

SocialCareCaseViewerApi/V1/Controllers/SubmissionController.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Produces("application/json")]
+    [ApiVersion("1.0")]
+    public class SubmissionController : BaseController
+    {
+        private readonly ISubmissionsUseCase _submissionsUseCase;
+
+        public SubmissionController(ISubmissionsUseCase submissionsUseCase)
+        {
+            _submissionsUseCase = submissionsUseCase;
+        }
```
chubberlisk, line 21: What do we think about `CaseSubmissionController` and `CaseSubmissionUseCase`? For me as someone with no context, submission is quite vague. We're using `CaseSubmissionResponse` etc., unless they'll be other types of "submissions"?

chubberlisk (commented):
I've tried to review the code as much as possible but tbh I struggled with knowing whether the actual case submission is the way to go and how that works i.e. the request passes one `socialCareId` but then multiple residents are returned. 🤔 I think I just lack context around this new feature and MongoDB. Happy to jump onto a call or let someone else review as well!

SocialCareCaseViewerApi.Tests/V1/UseCase/SubmissionUseCaseTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            _mockMongoGateway.Setup(x => x.InsertRecord(It.IsAny<string>(), It.IsAny<CaseSubmission>())).Verifiable();
+
+            var response = _submissionsUseCase.ExecutePost(request);
+            var expectedResponse = TestHelpers.CreateCaseSubmission(SubmissionState.InProgress, response.CreatedAt, worker, resident, response.FormId).ToDomain().ToResponse();
+
+            // struggle to verify _mockMongoGateway.InsertRecord as we do not have access to the dynamically created CaseSubmission object inserted
+            _mockDatabaseGateway.Verify(x => x.GetWorkerByEmail(request.CreatedBy), Times.Once);
+            _mockDatabaseGateway.Verify(x => x.GetPersonByMosaicId(request.ResidentId), Times.Once);
+
+            response.Should().BeEquivalentTo(expectedResponse);
```
chubberlisk, line 47: As discussed, let's see if we can spy on the `InsertRecord`.

JohnFarrellDev: I update the SubmissionsUseCase#ExecutePost function to return both the response object but also the infrastructure CaseSubmission object, this has allowed InsertRecord to be spied on.

I don't love this solution but it works, there's no real negative it just feels a little ugly.

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseSubmissionRequest.cs
JohnFarrellDev: We can use any name we want here, I like ResidentId as it distinguishes that the Id is about an individual resident.

A PersonId to me is a little vague about what the Person represents, could easily be an Id to lookup a Worker.

SocialCareCaseViewerApi/V1/Controllers/SubmissionController.cs
JohnFarrellDev: The Submissions are for forms in general that can be filled out for a resident.

How would you feel about `FormSubmissionController` and `FormSubmissionUseCase`?

chubberlisk: I like those suggestions! 👍

chubberlisk (approved):
Thanks for taking the time to go through this with me and addressing my comments! Feel free to ship after the rename. :ship:

PR #277, opened 2021-06-03
SCT-395 delete allocation records

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-395

Remove invalid allocation records

Added a plan for SQL updates to resolve the issue where non-existent person IDs are causing invalid allocation records to be in the database

Execute the plan and update the ticket


chubberlisk (commented):
We're missing the `.md` extension for the file so we can render as markdown. 👀

chubberlisk (approved):
Makes sense to me! Just need to add the file extension.

LBHTKarki:
> We're missing the `.md` extension for the file so we can render as markdown. 👀

Well spotted 👍 I've fixed that

PR #287, opened 2021-06-08
Bug(SCT-467): incorrect http response type from controller

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-467)

Unclear HTTP status codes are sent to the frontend. We are often sending a 500 when there is no actual server side error.

Updated return types.
Updated related tests.
In some places got would of error handling where `null` is not an exceptional circumstance. Just handle null.

#### _Checklist_


chubberlisk (approved):
Nice, responses are clearer!

PR #289, opened 2021-06-08
feature(SCT-416): patch endpoint for editing answers

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-416)

We want the ability to save a submission which means updating some of the answers stored against a form and the ability to later retrieve the form from that saved state.

Created a new API endpoint FormSubmissionController#EditSubmissionAnswers, takes form submission id and step (step is which part of the form you're on) as path parameters. From the body takes worker email who is doing the edit and also the dictionary object which represents the questions and their associated values for that part of the form.

FormSubmissionsUseCase#UpdateAnswers which validates the worker given and form to be updated are real (if not throw error and controller return bad request). Then update the answers and also update the EditHistory. Save update to database and return updated form.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Controllers/FormSubmissionControllerTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Controllers/FormSubmissionControllerTests.cs
@@ -104,5 +104,54 @@ namespace SocialCareCaseViewerApi.Tests.V1.Controllers
             response.Should().NotBeNull();
             response?.StatusCode.Should().Be(404);
         }
+
+        [Test]
+        public void EditSubmissionAnswersReturns200AndUpdatedResponse()
+        {
+            var createdSubmission = TestHelpers.CreateCaseSubmission().ToDomain().ToResponse();
+            const string stepId = "1";
+            var updateFormSubmissionAnswersRequest = TestHelpers.CreateUpdateFormSubmissionAnswersRequest();
+            _submissionsUseCaseMock.Setup(x => x.UpdateAnswers(createdSubmission.SubmissionId, stepId, updateFormSubmissionAnswersRequest)).Returns(createdSubmission);
+
+            var response = _formSubmissionController.EditSubmissionAnswers(createdSubmission.SubmissionId, stepId, updateFormSubmissionAnswersRequest) as ObjectResult;
+
+            response.Should().NotBeNull();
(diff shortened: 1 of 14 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateFormSubmissionAnswersRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public class UpdateFormSubmissionAnswersRequest
+    {
+        [FromBody]
+        [JsonPropertyName("editedBy")]
+        public string EditedBy { get; set; } = null!;
+
+        [FromBody]
+        [JsonPropertyName("stepAnswers")]
+        public Dictionary<string, object>
```
JohnFarrellDev, line 18: I'd love a way to represent this without using the object as the value.

Essentially this dictionary represents the questions and answers for a section of a form.
The question id acts as the key (typeof string) which is fine.
The answer can come in varying types from the [frontend](https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/data/flexibleForms/forms.types.ts#L58).

The issue is the type could be a string, a list of strings, or even a list of dictionaries that map a key of type string to either a string or list of strings.

Using the value type as an object works as a "catch all" for these types but unintentionally allows basically any type. Maybe this is ok because we don't ever expect this to be called except for by the frontend (if people aren't being sneaky). However, it is worth noting that the answers that map to a question id could be of an invalid type.

SocialCareCaseViewerApi.Tests/V1/Controllers/FormSubmissionControllerTests.cs
```diff
@@ -104,5 +104,54 @@ public void GetSubmissionByIdReturns404WhenACaseIsNotFound()
             response.Should().NotBeNull();
             response?.StatusCode.Should().Be(404);
         }
+
+        [Test]
```
Miles-Alford, line 108: Should there be a test for when a request is invalid and a 400 is returned?

JohnFarrellDev: Yes, good catch :+1: (test added)

Miles-Alford (approved):
👍  Looks good!

PR #291, opened 2021-06-08
Feature(SCT-348) PATCH endpoint for finishing a submission

Miles-Alford:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=59&modal=detail&selectedIssue=SCT-348)

We want the ability to finish a submission and store the submission in its finished state within the database

A new API endpoint called `FinishSubmission` has been added to the `FormSubmissionController` which takes in a `submissionID` from the path and a request boundary object called `FinishCaseSubmissionRequest` which passes the user's email who is finishing the submission from the body of the request.

The usecase has a new function called `ExecuteFinishSubmission` which validates that the worker and the submission which is to be updated both exist in the respective databases and then updates the specific submissions `SubmissionState` to be submitted and updates the history. No content is returned from this.

#### _Checklist_


SocialCareCaseViewerApi/V1/Boundary/Requests/FinishCaseSubmissionRequest.cs
```diff
@@ -0,0 +1,21 @@
+using System.Text.Json.Serialization;
+using FluentValidation;
+
```
JohnFarrellDev, line 3: you could add `#nullable enable` here.

https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/

It will give us that benefit that unless we define a property as null it can not be set to null (or the compiler will complain)

JohnFarrellDev (approved):
Really nice work, well done :)

PR #293, opened 2021-06-11
Fix: Make worker allocations null when executing upon a submission endpoint

Miles-Alford:
## Describe this PR

To avoid the circular reference exceptions some users with allocations receive when trying to create a new form submission, when a worker is retrieved from the database to be attached to the submission their allocations are set to null.
Both a workers `WorkerTeams` and `Allocations` properties are set to null before a submission is created/updated or finished

#### _Checklist_


chubberlisk (commented):
What do you think about adding tests to cover this by setting up a worker with `WorkTeams` and `Allocations` and then expecting that these attributes are set to null in the `caseSubmission`?

PR #294, opened 2021-06-11
Fix(SCT-508): Save the decision to disclose a warning note review

Renny4Real:
Previously, we were not retrieving nor saving a value for `DisclosedWithIndividual` on a warning note review.

In the frontend, a form element was added that would ask a user reviewing an ADULT warning note only, if they had disclosed the review to the individual.

When this form is submitted, the frontend would be making a PATCH request that would include a `DisclosedWithIndividual` value.

This change updates the service API to receive this updated request and process it.

#### _Checklist_

Test PATCH endpoint works and integrates with the frontend in Staging

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -1083,8 +1083,8 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         {
             var warningNote = TestHelpers.CreateWarningNote();
             var firstReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
-            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
-            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
+            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, true);
+            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, false);
 
             DatabaseContext.WarningNotes.Add(warningNote);
             DatabaseContext.WarningNoteReview.Add(firstReview);
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -1083,8 +1083,8 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         {
             var warningNote = TestHelpers.CreateWarningNote();
             var firstReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
-            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, true);
-            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, false);
+            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, disclosedWithIndividual: true);
+            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, disclosedWithIndividual: false);
 
             DatabaseContext.WarningNotes.Add(warningNote);
             DatabaseContext.WarningNoteReview.Add(firstReview);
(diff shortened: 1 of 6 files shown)


chubberlisk (approved):

SocialCareCaseViewerApi/V1/Boundary/Requests/PatchWarningNoteRequest.cs
```diff
@@ -16,6 +16,7 @@ public class PatchWarningNoteRequest
         public string ReviewNotes { get; set; }
         public string ManagerName { get; set; }
         public DateTime? DiscussedWithManagerDate { get; set; }
+        public bool DisclosedWithIndividual { get; set; }
```
chubberlisk, line 19: Oooooo, actually just realised does this need to be nullable since you mentioned it's only for adults?

chrishutchinson: I was just about to ask a similar question. Is this adults only? If so, are we being explicit with the validation / errors so that it's required if the resident type is adult, but shouldn't be passed if the resident type is child?

Renny4Real: Ah, this is true. 👍🏾 
I'll check the frontend request for both. I think the request from a child form would not have a `DisclosedWithIndividual` value since it's not part of the child form. 
And if that is the case, we could update this to allow `DisclosedWithIndividual` to be nullable?

Renny4Real: @chubberlisk @chrishutchinson I've just checked both requests and realised that we [modify the request](https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/075772eb411318aa816b0704f34293c71a9f6026/pages/people/%5Bid%5D/warning-notes/%5BwarningNoteId%5D/%5B%5B...stepId%5D%5D.tsx#L37) before we send it to the API.

My guess is if `disclosedWithIndividual` was null, the code `disclosedWithIndividual: formData.disclosedWithIndividual === 'Yes',` would return `false`.

What are your thoughts on changing the code in the frontend to check for and allow null, or should this be more granular, and we should check that the person is an adult first and then add the `disclosedWithIndividual` value to the request?

chrishutchinson: My expectation for the input/output would be that the type would be something like this (modelled in typescript just for convenience, but the theory should apply):

```ts
type Data = 
  | {
    residentType: "adult";
    disclosedWithIndividual: boolean;
  }
  | {
    residentType: "child";
    disclosedWithIndividual: never;
  }
```

This is effectively saying:
* If the resident is an adult, the value of `disclosedWithIndividual` **must** be provided, and it **must** be a `boolean`
* If the resident is a child, the value of `disclosedWithIndividual` should **never** be provided

This is more representative of the actual data model.

In the database we may have to store it as a `null` value, but I would expect the API layout should enforce these sort of rules.

Renny4Real: Thanks @chrishutchinson , 🙌🏾 
I will update the backend (this PR) so that the `disclosedWithIndividual` is nullable and then update the frontend to deal with `disclosedWithIndividual` based on the person.

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -1082,8 +1083,8 @@ public void GetWarningNoteByIdReturnsReturnsMultipleReviewsWithTheWarningNote()
         {
             var warningNote = TestHelpers.CreateWarningNote();
             var firstReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
-            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
-            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id);
+            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, true);
+            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, false);
```
chubberlisk, line 1087: For readability, could we used named arguments here? I think that's the right syntax... :sweat_smile:

```suggestion
            var secondReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, disclosedWithIndividual: true);
            var thirdReview = TestHelpers.CreateWarningNoteReview(warningNote.Id, disclosedWithIndividual: false);
```

chubberlisk (approved):

PR #300, opened 2021-06-15
Refactor/remove should not be null checks

JohnFarrellDev:
## Link to JIRA ticket

No ticket required.

We have loads of tests that have a check `Should().NotBeNull()` then following test such as `Should().BeEquivalentTo(expectedResponse)`. The null check is pointless because the test doesn't care what the response isn't, it cares what it is and we already check it is equivalent to expectedResponse.

Delete all these pointless checks, make our tests tidier.

#### _Checklist_


JohnFarrellDev:
:cowboy_hat_face:

PR #302, opened 2021-06-16
Refactor(): Make SubmissionState API response more (human) readable

JohnFarrellDev:
## Link to JIRA ticket

No ticket required, small change to make response nicer from API.

We store SubmissionState of a CaseSubmission in an enum which is logical, however we want a nicer human readable output instead of ints. Simply map enum value to appropriate string.

#### _Checklist_


jhackett1 (approved):
we like it

PR #304, opened 2021-06-16
Remove PostgreSQL import lambda from the diagram

LBHTKarki:
## Describe this PR

Remove obsolete PostgreSQL import Lambda from the diagram

Diagram is out of date now that we have removed the Lambda from the solution.

Updated the diagram

#### _Checklist_


chubberlisk (approved):
💯

PR #308, opened 2021-06-17
Feature(SCT-459): supplement cases api return completed case submissions

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-459)

We want when going to a resident's page to see any submitted CaseSubmissions returned as part of their "records".

Amended CaseRecordsUseCase to make a request to MongoGateway to retrieve all CaseSubmissions that belong to the resident if they are in the submitted state. Transform this data into "CareCaseData" object so it can be returned via the original API.

#### _Checklist_


JohnFarrellDev:
> do we expect any performance issues from adding a another data source for this endpoint to call?

It should be okay, can check in staging and if we see any performance issues look to make some improvements.

PR #310, opened 2021-06-17
refactor(SCT-588): rename relationships classes

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-588

### *What is the problem we're trying to solve*

We want to refactor the current `residents/{personId}/relationships` API endpoint to change the response with more granular relationship types. In order to do this, we've decided that the best way to go about this without breaking current functionality to:

1. Rename current relationships classes to have `V1` in it, i.e. gateway, use case, etc.
2. Deploy to prod
3. Duplicate the current relationship API endpoint and rename one to `residents/{personId}/relationships-v1`
4. Deploy to prod
5. Update frontend to use `residents/{personId}/relationships-v1`
6. Deploy to prod
7. Add API code for how we want to respond with new relationships types
8. Deploy to prod
9. Update frontend to switch over to use `residents/{personId}/relationships` i.e. endpoint with new response
10. Delete v1 relationships code

### *What changes have we introduced*

This PR does step 1 and renames relationships classes to have `V1` in it.

#### _Checklist_

- [ ] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature
- [ ] Added tests to cover all new production code
- [ ] Added comments to the README or updated relevant documentation _(add link to documentation)_, where necessary.
- [x] Checked all code for possible refactoring
- [x] Code pipeline builds correctly

### *Follow up actions after merging PR*

Deploy to prod and then duplicate endpoint.

PR #311, opened 2021-06-18
refactor(SCT-589): refactor get relationships endpoint into new RelationshipController

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-589

We want to duplicate the current get relationships API endpoint (`residents/{personId}/relationships`) as we need to completely change the response without breaking current functionality.

This PR adds a new controller called `RelationshipController` and then moves the current relationships endpoint into there. Decided to split this first bit off so it's easier to review and we can get this deployed alongside the renaming in #310.

Duplicate the relationships endpoint but with different path (`residents/{personId}/relationships-v1`)

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
@@ -17,8 +17,8 @@ namespace SocialCareCaseViewerApi.Tests.V1.Controllers
     {
         private RelationshipController _classUnderTest;
         private Mock<IRelationshipsV1UseCase> _mockRelationshipsUseCase;
-        private Fixture _fixture;
-        private Faker _faker;
+        private readonly Fixture _fixture = new Fixture();
+        private readonly Faker _faker = new Faker();
 
         [SetUp]
         public void SetUp()
@@ -26,9 +26,6 @@ namespace SocialCareCaseViewerApi.Tests.V1.Controllers
             _mockRelationshipsUseCase = new Mock<IRelationshipsV1UseCase>();
 
             _classUnderTest = new RelationshipController(_mockRelationshipsUseCase.Object);
-
-            _fixture = new Fixture();
-            _faker = new Faker();
         }
 
         [Test]
--- a/SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
@@ -1,0 +1,74 @@
+using AutoFixture;
+using Bogus;
+using FluentAssertions;
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+using NUnit.Framework;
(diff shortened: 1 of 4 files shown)


LBHTKarki (approved):

SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        private Faker _faker;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _mockRelationshipsUseCase = new Mock<IRelationshipsV1UseCase>();
+
+            _classUnderTest = new RelationshipController(_mockRelationshipsUseCase.Object);
+
+            _fixture = new Fixture();
```
JohnFarrellDev, line 30: You could move the `new Fixture()` to where the field is declared and it can then be made `readonly`, [SetUp] is ran before every test so before each test we make a new Fixture object and assign it to _fixture.

`private readonly Fixture _ fixture = new Fixture()`

^ same should work for _faker.

SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var request = new ListRelationshipsV1Request() { PersonId = _faker.Random.Long() };
+
+            var listRelationShipsResponse = _fixture.Create<ListRelationshipsV1Response>();
+
+            _mockRelationshipsUseCase.Setup(x => x.ExecuteGet(It.IsAny<ListRelationshipsV1Request>())).Returns(listRelationShipsResponse);
+
+            var response = _classUnderTest.ListRelationships(request) as ObjectResult;
+
+            response?.Value.Should().BeOfType<ListRelationshipsV1Response>();
```
JohnFarrellDev, line 70: Is this check required?

I imagine `BeEquivalentTo` would fail if the types don't match.

JohnFarrellDev:
Good clean code and I love that you removed an API endpoint out of that SocialCareCaseViewerApi class.

JohnFarrellDev (approved):
:+1:

SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
chubberlisk: Just tested, all good! This is nicer as it makes the `SetUp` method leaner. 👍

SocialCareCaseViewerApi.Tests/V1/Controllers/RelationshipControllerTests.cs
chubberlisk: I wasn't sure tbh but just discovered `BeEquivalentTo` doesn't care about the type, see below! Whether or not we care about the type if the data and structure is what we expect, I'm not sure.

image

mapleleafjack (approved):

PR #317, opened 2021-06-21
feat(SCT-412): add DatabaseGateway#GetPersonWithPersonalRelationshipsByPersonId 

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-412

We've set everything up to allow us to change the response for getting relationships, now it's time to write the code to do dis!

This PR adds a new method in the `DatabaseGateway` to get a person with their personal relationships. For each personal relationship, it will also retrieve the details of the other person, the description of the relationship type and its details
e.g. "Emergency contact".

Some noticeable things:

- A completely new and separate test file is added even though it's part of `DatabaseGateway` as `DatabaseGateway.Tests.cs` is BIG -> `SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs` 
- A completely new and separate test helper is added that is just for personal relationships as `TestHelpers` is BIG -> `SocialCareCaseViewerApi.Tests/V1/Helpers/PersonalRelationshipsHelper.cs`

#### _Checklist_

Create the use case to map the response from the gateway.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
@@ -71,13 +71,23 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
         }
 
         [Test]
-        public void WhenThereIsARelationshipHasAnEndDateReturnsEmptyListForPersonalRelationships()
+        public void WhenIncludeEndedRelationshipsIsFalseReturnsEmptyListForPersonalRelationships()
         {
             var (person, _, _, _, _) = PersonalRelationshipsHelper.SavePersonWithPersonalRelationshipToDatabase(DatabaseContext, hasEnded: true);
 
-            var response = _databaseGateway.GetPersonWithPersonalRelationshipsByPersonId(person.Id);
+            var response = _databaseGateway.GetPersonWithPersonalRelationshipsByPersonId(person.Id, includeEndedRelationships: false);
 
             response.PersonalRelationships.Should().BeEmpty();
+        }
+
+        [Test]
+        public void WhenIncludeEndedRelationshipsIsTrueReturnsPersonalRelationships()
+        {
+            var (person, _, _, _, _) = PersonalRelationshipsHelper.SavePersonWithPersonalRelationshipToDatabase(DatabaseContext, hasEnded: true);
+
+            var response = _databaseGateway.GetPersonWithPersonalRelationshipsByPersonId(person.Id, includeEndedRelationships: true);
+
+            response.PersonalRelationships.Should().NotBeEmpty();
         }
 
         [Test]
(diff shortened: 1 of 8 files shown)


SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
```diff
@@ -0,0 +1,104 @@
...
+using FluentAssertions;
+using MongoDB.Driver;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Gateways;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
+{
+    [TestFixture]
```
LBHTKarki, line 11: Should there be a test for a case where the person with the given id doesn't exist in the database?

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -880,6 +880,18 @@ public List<Person> GetPersonsByListOfIds(List<long> ids)
...
         }
 
+        public Person GetPersonWithPersonalRelationshipsByPersonId(long personId)
+        {
+            var personWithRelationships = _databaseContext
+                .Persons
+                .Include(person => person.PersonalRelationships)
+                .FirstOrDefault(p => p.Id == personId);
+
+            personWithRelationships.PersonalRelationships = personWithRelationships.PersonalRelationships.Where(pr => pr.EndDate == null).ToList();
```
LBHTKarki, line 890: This will throw an exception if the person doesn't exist

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
chubberlisk: Definitely! 🤦‍♀️

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
LBHTKarki, line 890: I wonder if there should be a parameter that would determine whether only active relationships should be returned?

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
chubberlisk: Added `if (personWithRelationships == null) return null;` check alongside a test.

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
chubberlisk: Related to your above comment, I've added a check before this to ensure person isn't `null` so this doesn't happen anymore.

chubberlisk: We can add one and default it to only active ones for now and if we ever need it we can always switch it in the use case.

mapleleafjack (approved):

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
chubberlisk: I've added a `includeEndedRelationships` to the method and set it `false` as the default.

LBHTKarki (approved):

PR #318, opened 2021-06-21
Feat(SCT-630): Add list endpoint for incomplete submissions

Renny4Real:
The new form builder created for the application deals with completing a form by saving the current state of the submission as a status property in MongoDB.

This adds a `LIST` endpoint that, when called, would return a list of all **in-progress** submissions only.

#### _Checklist_

- [X] Added comments to the README or updated, relevant documentation _(add a link to documentation)_, where necessary.

### *ACTIONS AFTER MERGING PR*

- Test the endpoint in staging

### DIFF
--- a/SocialCareCaseViewerApi/V1/Controllers/FormSubmissionController.cs
+++ b/SocialCareCaseViewerApi/V1/Controllers/FormSubmissionController.cs
@@ -1,8 +1,10 @@
+using System.Collections.Generic;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using SocialCareCaseViewerApi.V1.Boundary.Requests;
 using SocialCareCaseViewerApi.V1.Boundary.Response;
 using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.Infrastructure;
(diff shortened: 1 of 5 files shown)


SocialCareCaseViewerApi/V1/Controllers/FormSubmissionController.cs
```diff
@@ -40,6 +42,20 @@ public IActionResult GetSubmissionById(string submissionId)
             return Ok(submission);
         }
 
+        /// <summary>
+        /// Lists all in-progress case submissions
+        /// </summary>
+        /// <response code="200">Success. Returns a list of any in progress case submissions</response>
+        [ProducesResponseType(typeof(List<CaseSubmissionResponse>), StatusCodes.Status200OK)]
+        [HttpGet]
+        [Route("status/in-progress")]
```
jhackett1, line 51: probably this can just be `/submissions` to stick to REST conventions - no need for the extra path bits

at most we might want to use a query string parameter like `?status=` to filter more intelligently down the line, but not needed for now

Renny4Real: 👍🏾 I can update this

jhackett1 (approved):
exactly what we need

PR #322, opened 2021-06-23
Bugfix: add loop to remove circular reference for allocation worker

Miles-Alford:
## Link to JIRA ticket

None

To try and solve the bug related to the mongoDB circular reference issue when trying to post a new submission, A for loop has been added which makes the person associated with an allocation null so that it does not loop around and keep referencing itself

Add a for loop which makes the person attribute of a allocation null

#### _Checklist_


LBHTKarki (changes requested):
For clarity, please update the description so that it says this removes the person, rather than worker, reference associated with the allocation.

LBHTKarki (approved):

PR #345, opened 2021-06-30
feat(SCT-263): add database gateway methods for adding personal relationships

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-263

We want to be able to add a personal relationship. 

This PR adds two new `DatabaseGateway` methods:

- ` GetPersonalRelationshipTypeByDescription(string description)`
-  `CreatePersonalRelationship(CreatePersonalRelationshipRequest request)`

which will be used to support the new use case for creating a personal relationship that will handle:

- checking that the person and other person exist using _existing_ DB gateway method
- checking that the personal relationship type exists using _new_ DB gateway method
- checking that the personal relationship with the same type doesn't already exist using _existing_ DB gateway method
- creating the personal relationship using _new_ DB gateway method
- creating the inverse personal relationship using _new_ DB gateway method

#### _Checklist_

Create new use case for adding personal relationship.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
@@ -36,11 +36,11 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
             _databaseGateway.CreatePersonalRelationship(request);
 
             var personalRelationship = DatabaseContext.PersonalRelationships.FirstOrDefault();
-            personalRelationship.PersonId.Should().Be(request.PersonId);
-            personalRelationship.OtherPersonId.Should().Be(request.OtherPersonId);
-            personalRelationship.TypeId.Should().Be(request.TypeId);
-            personalRelationship.IsMainCarer.Should().Be(request.IsMainCarer);
-            personalRelationship.IsInformalCarer.Should().Be(request.IsInformalCarer);
+            personalRelationship?.PersonId.Should().Be(request.PersonId);
+            personalRelationship?.OtherPersonId.Should().Be(request.OtherPersonId);
+            personalRelationship?.TypeId.Should().Be(request.TypeId);
+            personalRelationship?.IsMainCarer.Should().Be(request.IsMainCarer);
+            personalRelationship?.IsInformalCarer.Should().Be(request.IsInformalCarer);
         }
 
         [Test]
@@ -58,7 +58,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
             _databaseGateway.CreatePersonalRelationship(request);
 
             var personalRelationship = DatabaseContext.PersonalRelationships.FirstOrDefault();
-            personalRelationship.StartDate.Should().Be(fakeTime);
+            personalRelationship?.StartDate.Should().Be(fakeTime);
         }
 
         [Test]
@@ -73,8 +73,8 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
             var response = _databaseGateway.CreatePersonalRelationship(request);
 
             var details = DatabaseContext.PersonalRelationshipDetails.FirstOrDefault();
-            details.PersonalRelationshipId.Should().Be(response.Id);
-            details.Details.Should().Be(request.Details);
+            details?.PersonalRelationshipId.Should().Be(response.Id);
+            details?.Details.Should().Be(request.Details);
         }
     }
 }
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Linq;
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using Microsoft.EntityFrameworkCore;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
+{
+    [TestFixture]
+    public class CreatePersonalRelationshipTests : DatabaseTests
+    {
+        private DatabaseGateway _databaseGateway;
+        private Mock<IProcessDataGateway> _mockProcessDataGateway = new Mock<IProcessDataGateway>();
+        private Mock<ISystemTime> _mockSystemTime = new Mock<ISystemTime>();
+
+        [SetUp]
+        public void Setup()
+        {
+            _databaseGateway = new DatabaseGateway(DatabaseContext, _mockProcessDataGateway.Object, _mockSystemTime.Object);
+            DatabaseContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
+        }
+
+        [Test]
+        public void CreatesAPersonalRelationship()
+        {
+            var (person, otherPerson) = PersonalRelationshipsHelper.SavePersonAndOtherPersonToDatabase(DatabaseContext);
+            var type = DatabaseContext.PersonalRelationshipTypes.FirstOrDefault(prt => prt.Description == "parent");
+            var request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(
+                person.Id, otherPerson.Id, type.Id, type.Description
+            );
+
+            _databaseGateway.CreatePersonalRelationship(request);
+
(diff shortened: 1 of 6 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/CreatePersonalRelationshipRequest.cs
```diff
@@ -16,6 +16,8 @@ public class CreatePersonalRelationshipRequest
         [JsonPropertyName("type")]
         public string Type { get; set; } = null!;
 
+        public long TypeId { get; set; }
```
JohnFarrellDev, line 19: We might want to utilise `[JsonPropertyName("propertyName")] ` here, keeps it consistent with the rest of the class and makes it very easy to update the expected JSON body key without changing any variable names.

JohnFarrellDev: Also just to confirm, are we always expecting TypeId to be provided? If not it should be made nullable.

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -905,6 +905,34 @@ public Person GetPersonWithPersonalRelationshipsByPersonId(long personId, bool i
             return personWithRelationships;
         }
 
+        public PersonalRelationshipType GetPersonalRelationshipTypeByDescription(string description)
+        {
+            return _databaseContext.PersonalRelationshipTypes
+                .FirstOrDefault(prt => prt.Description == description);
```
JohnFarrellDev, line 911: should we care about case? You might want to call `.ToLower()` against both strings.

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonalRelationshipTypeByDescriptionTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
+{
+    [TestFixture]
+    public class GetPersonalRelationshipTypeByDescriptionTests : DatabaseTests
+    {
+        private DatabaseGateway _databaseGateway;
+        private Mock<IProcessDataGateway> _mockProcessDataGateway = new Mock<IProcessDataGateway>();
+        private Mock<ISystemTime> _mockSystemTime = new Mock<ISystemTime>();
```
JohnFarrellDev, line 16: These could be made `readonly`

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonalRelationshipTypeByDescriptionTests.cs
```diff
@@ -0,0 +1,46 @@
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.V1.Helpers;
+using Microsoft.EntityFrameworkCore;
+using System;
```
JohnFarrellDev, line 7: this once can be removed

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonalRelationshipTypeByDescriptionTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            response.Should().BeNull();
+        }
+
+        [Test]
+        [TestCase("parent")]
+        [TestCase("child")]
+        [TestCase("friend")]
+        public void WhenMatchingDescriptionReturnsPersonalRelationshipType(string description)
+        {
+            var response = _databaseGateway.GetPersonalRelationshipTypeByDescription(description);
```
JohnFarrellDev, line 39: Just to confirm my understanding `parent`, `child` and `friend` are all relationship types we have stored in the database by default.

This is why this test is able to call `_databaseGateway.GetPersonalRelationshipTypeByDescription` before adding anything to the db first?

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
+{
+    [TestFixture]
+    public class CreatePersonalRelationshipTests : DatabaseTests
+    {
+        private DatabaseGateway _databaseGateway;
+        private Mock<IProcessDataGateway> _mockProcessDataGateway = new Mock<IProcessDataGateway>();
+        private Mock<ISystemTime> _mockSystemTime = new Mock<ISystemTime>();
```
JohnFarrellDev, line 18: can be made `readonly`

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            var (person, otherPerson) = PersonalRelationshipsHelper.SavePersonAndOtherPersonToDatabase(DatabaseContext);
+            var type = DatabaseContext.PersonalRelationshipTypes.FirstOrDefault(prt => prt.Description == "parent");
+            var request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(
+                person.Id, otherPerson.Id, type.Id, type.Description
+            );
+
+            _databaseGateway.CreatePersonalRelationship(request);
+
+            var personalRelationship = DatabaseContext.PersonalRelationships.FirstOrDefault();
+            personalRelationship.PersonId.Should().Be(request.PersonId);
```
JohnFarrellDev, line 39: for checks like this we might want to do

`personalRelationship?.PersonId.Should().Be(request.PersonId);`

This way if personalRelationships is null instead of `NullReferenceException` being thrown we will have the test fail because `null` is not equal to `request.PersonId`.

JohnFarrellDev: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

JohnFarrellDev (approved):
Trixie reckons it's good to go.

IMG_20210630_134452

SocialCareCaseViewerApi/V1/Boundary/Requests/CreatePersonalRelationshipRequest.cs
chubberlisk: So `TypeId` won't be provided as part of the frontend request, instead it will be added by the use case for the gateway to use. Do you think that too cheeky? Basically wanted to avoid creating another class for the gateway method parameter... 👀 

But you're right we need to make it nullable, thanks!

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
chubberlisk: I guess we shouldn't really. Will add dis!

SocialCareCaseViewerApi/V1/Boundary/Requests/CreatePersonalRelationshipRequest.cs
JohnFarrellDev: that's fine, just add the `[jsonIgnore]` attribute

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
chubberlisk: Done in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/345/commits/e07b2c23bca6bb074d9c0ffb99407cd19731d6e1!

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonalRelationshipTypeByDescriptionTests.cs
chubberlisk: Yup, that's exactly right - it's been added to the `schema.sql` which means we don't have to seed the database with personal relationship types.

https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/5e5216b89ec53fa1e7b6a26119119e74a6faf083/database/schema.sql#L431-L499

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
chubberlisk: Ooooooooo, I really like this! Thanks for the suggestion! 💯

PR #346, opened 2021-06-30
feat(SCT-263): add new use case for adding a personal relationship

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-263

We want to be able to add a personal relationship. In #345, we added methods in the `DatabaseGateway` to support this new use case.

This PR adds a new use case: `PersonalRelationship#ExecutePost` that will ensure that:

- the provided `personId` does exist
- the provided `otherPersonId` does exist
- the provided `description` of type does exist
- the provided relationship doesn't already exist for the same person with the same type

It will then create a new personal relationship for the provided person and create the inverse relationship.

#### _Checklist_

Create the API endpoint: `api/v1/relationships/personal`.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
@@ -1,0 +1,156 @@
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.V1.Infrastructure;
+using SocialCareCaseViewerApi.V1.UseCase;
+using System.Collections.Generic;
+using System;
+
+namespace SocialCareCaseViewerApi.Tests.V1.UseCase.Relationships
+{
+    [TestFixture]
+    public class PersonalRelationshipsExecutePostUseCaseTests
+    {
+        private Mock<IDatabaseGateway> _mockDatabaseGateway;
+        private PersonalRelationshipsUseCase _personalRelationshipsUseCase;
+        private CreatePersonalRelationshipRequest _request;
+        private Person _person;
+        private Person _otherPerson;
+        private readonly PersonalRelationshipType _typeInRequest = PersonalRelationshipsHelper.CreatePersonalRelationshipType("friend");
+
+        [SetUp]
+        public void SetUp()
+        {
+            _mockDatabaseGateway = new Mock<IDatabaseGateway>();
+            _personalRelationshipsUseCase = new PersonalRelationshipsUseCase(_mockDatabaseGateway.Object);
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonalRelationshipTypeByDescription(_typeInRequest.Description))
+                .Returns(_typeInRequest);
+
+            var typeInExistingRelationship = PersonalRelationshipsHelper.CreatePersonalRelationshipType("partner");
+            _mockDatabaseGateway.Setup(x => x.GetPersonalRelationshipTypeByDescription(typeInExistingRelationship.Description))
+                .Returns(typeInExistingRelationship);
+
+            _request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(type: _typeInRequest.Description);
+
+            _person = TestHelpers.CreatePerson((int) _request.PersonId);
+            _otherPerson = TestHelpers.CreatePerson((int) _request.OtherPersonId);
+
+            var personalRelationship = PersonalRelationshipsHelper.CreatePersonalRelationship(_person, _otherPerson, typeInExistingRelationship);
+            _person.PersonalRelationships = new List<PersonalRelationship>() { personalRelationship };
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonsByListOfIds(It.IsAny<List<long>>()))
+                .Returns(new List<Person>() { _person, _otherPerson });
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonWithPersonalRelationshipsByPersonId(It.IsAny<long>(), It.IsAny<bool>()))
+                .Returns(_person);
+
+            _mockDatabaseGateway.Setup(x => x.CreatePersonalRelationship(It.IsAny<CreatePersonalRelationshipRequest>()))
+                .Returns(personalRelationship);
+        }
+
+        [Test]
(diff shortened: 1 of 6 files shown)


SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
```diff
@@ -0,0 +1,52 @@
...
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.V1.UseCase.Interfaces;
+
+namespace SocialCareCaseViewerApi.V1.UseCase
+{
+    public class PersonalRelationshipsUseCase : IPersonalRelationshipsUseCase
+    {
+        private IDatabaseGateway _databaseGateway;
```
JohnFarrellDev, line 11: can be `readonly`

SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
```diff
@@ -0,0 +1,52 @@
...
+            var typeDoesNotExist = type == null;
+            if (typeDoesNotExist) throw new PersonalRelationshipTypeNotFoundException($"\"type\" with \"{request.Type}\" was not found.");
+
+            var personWithPersonalRelationships = _databaseGateway.GetPersonWithPersonalRelationshipsByPersonId(request.PersonId);
+
+            var personalRelationships = personWithPersonalRelationships.PersonalRelationships;
+            var personalRelationshipAlreadyExists = personalRelationships.Find(pr => pr.OtherPersonId == request.OtherPersonId && pr.Type.Description == request.Type) != null;
+            if (personalRelationshipAlreadyExists) throw new PersonalRelationshipAlreadyExistsException($"Personal relationship with \"type\" of \"{request.Type}\" already exists.");
+
+            _databaseGateway.CreatePersonalRelationship(request);
```
JohnFarrellDev, line 39: If the request.TypeId is null the CreatePersonalRelationship function will throw an exception I think when it tries to cast null value to type long.

SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            _databaseGateway.CreatePersonalRelationship(new CreatePersonalRelationshipRequest()
+            {
+                PersonId = request.OtherPersonId,
+                OtherPersonId = request.PersonId,
+                TypeId = type.InverseTypeId,
+                IsMainCarer = null,
+                IsInformalCarer = null,
+                Details = null
```
JohnFarrellDev, line 48: The CreatePersonalRelationshipRequest object states these values can't be null.

You could make them nullable but keep the FluentValidation checks to ensure the actual API request validates they aren't null. This way you know the request actually has assigned the values but can also set the value as null directly yourself without compiler warnings.

SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            _personalRelationshipsUseCase.ExecutePost(_request);
+
+            _mockDatabaseGateway.Verify(gateway => gateway.GetPersonalRelationshipTypeByDescription(_request.Type));
+        }
+
+        [Test]
+        public void WhenTypeDoesNotExistThrowsPersonalRelationshipTypeNotFoundException()
+        {
+            _mockDatabaseGateway.Setup(x => x.GetPersonalRelationshipTypeByDescription(It.IsAny<string>()))
+                .Returns((PersonalRelationshipType) null);
```
JohnFarrellDev, line 100: If you want your mock to return null you can just leave off the `.Returns(...)`

SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,156 @@
+using FluentAssertions;
```
JohnFarrellDev, line 1: Good very comprehensive tests :)

JohnFarrellDev (approved):
:rocket:

SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
chubberlisk: Not going to lie, I enjoyed TDDing this very much! 🕺

SocialCareCaseViewerApi.Tests/V1/UseCase/PersonalRelationships/PersonalRelationshipsExecutePostUseCaseTests.cs
chubberlisk: Ahhhh that's true! But I think I'd prefer to keep it and have the test be explicit about it for readability and clarity.

SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
chubberlisk: I need to get VS Code to pick these up! 🤦

SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
chubberlisk: m8, good spot! I completely forgot I needed to do this... I did it for the inverse but not for this one. 🤦

SocialCareCaseViewerApi/V1/UseCase/PersonalRelationshipsUseCase.cs
chubberlisk: That makes sense, I think I keep going confused by `null!`. Could you check that the latest commit is correct please?

JohnFarrellDev: So if something is nullable (got the ? after the type) we don't need `null!`.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

My understanding is saying `null!` is basically telling the compiler I know this isn't meant to be null but I know right now I'm not assigning this a value so for now it is null and that's ok.

chubberlisk: Ahhh okay, I got! Your explanation is much clearer than the docs for me. 🙇‍♀️

Will remove `null!`.

PR #348, opened 2021-07-01
Renamed file which blocked cloning on Windows

martapederiva:
## Describe this PR

When trying to clone this repository on Windows, one file with < and > was blocking the cloning.
The file was renamed deleting the offending characters and the filename was also changed in the README.

Cloning repo on Windows.

None


JohnFarrellDev (approved):
:) Whoopsie, you could change the target branch to `master` if you want It's not like the code changes that require testing on a staging.

PR #350, opened 2021-07-02
Add manual update for fixing types of personal relationships

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-764

When we transformed the historic personal relationships, we incorrectly set the `fk_personal_relationship_type_id`.

This PR adds the manual update for fixing types of personal relationships.

Run the plan!

### DIFF
--- a/database/manual-updates/2021-07-02_1-fix-types-of-historic-personal-relationship.md
+++ b/database/manual-updates/2021-07-02_1-fix-types-of-historic-personal-relationship.md
@@ -1,0 +1,56 @@
+# Fix types of historic personal relationships
+
+## The problem we're trying to solve
+
+In the [manual update to transform the historic personal relationships](2021-06-16_2-transform-historic-personal-relationships.md),
+we incorrectly set the `fk_personal_relationship_type_id`.
+
+## Justification for doing a manual update
+
+This is the simplest and quickest way to transform the data, using DMS for this
+is overkill.
+
+## The plan
+
+1. Make a backup of the `dbo.sccv_personal_relationship` table
+2. Run an update statement to set the `fk_personal_relationship_type_id` to the
+   `inverse_type_id` of the current value for a relationship for the person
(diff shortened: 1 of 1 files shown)


database/manual-updates/2021-07-02_1-fix-types-of-historic-personal-relationship.md
```diff
@@ -0,0 +1,57 @@
...
+SET fk_personal_relationship_type_id = prt.inverse_type_id
+FROM dbo.sccv_personal_relationship_type prt
+WHERE pr.fk_personal_relationship_type_id = prt.id AND pr.id = <OTHER_PERSONAL_RELATIONSHIP_ID>;
+
+-- Then, update all relationships
+-- Update personal relationships
+UPDATE dbo.sccv_personal_relationship pr
+SET fk_personal_relationship_type_id = prt.inverse_type_id
+FROM dbo.sccv_personal_relationship_type prt
+WHERE pr.fk_personal_relationship_type_id = prt.id AND NOT (pr.id = <PERSONAL_RELATIONSHIP_ID> OR pr.id = <OTHER_PERSONAL_RELATIONSHIP_ID>);
```
LBHTKarki, line 52: Should that AND NOT... statement be left out perhaps if updating all records?

chubberlisk: So the reason we need the `AND NOT` is because we were thinking it might be a good idea to test with one relationship first and so we don't want to inverse the ones we have already done again otherwise they would go back to being incorrect.

chubberlisk: Will update the comment to be clearer!

LBHTKarki: That makes sense, I was thinking the test being separate somehow

LBHTKarki (approved):

PR #352, opened 2021-07-02
SCT-490 remove mosaic api dependency

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-490

This removes the dependency to Mosaic Resident Information platform API used in person search. This enables us to progress with new person search improvements a lot quicker since we don't have to worry about the platform API anymore.

- Copied over the gateway from the platform API keeping the logic and objects the same to ensure backwards compatibility
- Modified all tests to use our new helpers for consistency
- I had to tweak some of the existing helpers to make them more flexible when testing person data. No unrelated tests were modified because of this
- I didn't implement the post code validation since we are heading towards partial search for it later on. Front end has validation anyway so that'll take care of that for now
- I kept the boundary objects the same to ensure this can be deployed without frontend changes. Those objects can be updated later if necessary 
- There are known issues in the way the gateway logic works (person details vs address search), but I didn't want to touch that at this stage to ensure we don't introduce too many changes at the same time
- Migrated gateway methods have to be refactored further to tidy things up. For now the main thing is to get rid of the platform API dependency

#### _Checklist_


SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetAllResidentsTests.cs
```diff
@@ -0,0 +1,506 @@
...
+            var person1 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            var person2 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            var person3 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+
+            person3.DateOfBirth = person1.DateOfBirth;
+
+            DatabaseContext.Persons.AddRange(new List<Person> { person1, person2, person3 });
+            DatabaseContext.SaveChanges();
+
+            var listOfPersons = _classUnderTest.GetResidentsBySearchCriteria(cursor: 0, limit: 20, dateOfBirth: person1.DateOfBirth.Value.ToString("yyyy-MM-dd")).ToList(); //Format passed be FE
```
chubberlisk, line 313:
```suggestion
            var listOfPersons = _classUnderTest.GetResidentsBySearchCriteria(cursor: 0, limit: 20, dateOfBirth: person1.DateOfBirth.Value.ToString("yyyy-MM-dd")).ToList(); //Format passed by FE
```

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetAllResidentsTests.cs
```diff
@@ -0,0 +1,506 @@
...
+        public void WithNameAndPostCodeQueryParameterReturnsMatchingResident()
+        {
+            var person1 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            person1.FirstName = "ciasom";
+
+            var person2 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            person2.FirstName = "wrong name";
+
+            var person3 = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            person3.FirstName = "ciasom";
```
chubberlisk, line 349: Looks like `CreatePersonDatabaseEntity` takes in first name as a parameter, could this be refactored to use it?

chubberlisk (approved):

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetAllResidentsTests.cs
LBHTKarki: yes, forgot to tidy that up. I've updated it now

PR #353, opened 2021-07-05
feat(SCT-471): update isMainCarer and isInformalCarer to be nullable in API request

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-471

We've restricted `isMainCarer` and `isInformalCarer` to be required but they are times when it's not relevant to have both or either e.g. unborn child etc.

This PR removes the validation check for `isMainCarer` is not null and `isInformalCarer` is not null.

#### _Checklist_


LBHTKarki (approved):
@chubberlisk will this have an impact on the front end i.e. can it handle null values for this?

chubberlisk:
> @chubberlisk will this have an impact on the front end i.e. can it handle null values for this?

Yup, it will but I'm working on the frontend and updating it. 😉

PR #360, opened 2021-07-08
[Breaking changes] Deleted records handling

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-396

This is part of the duplicate records clean up work we are doing with the data insight team and the services.

Services have identified numerous person records that are in fact for the same individual. Services have identified those IDs and combined the submitted form records under one master ID. This is typically the original restored Mosaic ID, but in some cases it could be an ID created by mistake. We need to be able to delete those offending records to stop them causing confusion to users.

To complete the clean up task we need to be able to delete all person and related records from the tables. However because the ID mapping work is a manual process there's a risk that some person records that should remain are deleted by accident. For easy recovery from that situation this PR provides a method for soft deleting the person and other related records.

Soft delete is achieved by setting the `marked_for_deletion` flag to true for all records that should be deleted. **This means all person related tables have been updated with a new column**. Flags are set by SQL script that has test coverage to increase confidence that it's doing the right updates. 

To set the flag for appropriate records:

1. Import the data provided by the data insight team to the `dbo.sccv_person_record_to_be_deleted` table. Foreign keys are used to prevent invalid IDs getting imported
2. Run then SQL script from the test file wrapping it into transaction. That ensures all records get updated correctly

Once the script runs successfully it'll update the `dbo.sccv_deleted_person_record` table with records that show what IDs were deleted and what the new master record IDs are. This was requested by the data insight team for audit purposes.

This PR also introduces handling for the deleted records from the API point of view. Since the soft delete is only temporary i.e. those records will get deleted after short cool down period, the API updates are not meant to cover every available scenario. Updates are only made to end points that are involved in front end user trying to find a person. This should resolve the immediate issue where users can see deleted records and add further details against them. 

Following scenarios are covered :

- Person entity: stop users finding person records that are marked for deletion. This includes search and various GET end points involving person ID
- Relationship: stop users seeing relationships against deleted person records
- Allocations: it was agreed with the data team that all allocations against the deleted records are removed in advance, so there's no need to handle them in the code

To fully cover all end points and logic in the app would be a lot of work, but this should help to hide the records that are causing issues to users.

#### _Checklist_

- Create a ticket for hard deletion of records (postgreSQL and DocumentDB). Soft deletion of DocumentDB records is not part of this PR and can be looked into separately. That'll be part of the https://hackney.atlassian.net/browse/SCT-765
- Create a ticket for restoring records from soft delete
- Moving forward all new person related entities must be added to this script so they are covered by the process


database/schema.sql
```diff
@@ -500,3 +500,44 @@ update dbo.sccv_personal_relationship_type set inverse_type_id = (select id from
...
+ALTER TABLE dbo.sccv_warning_note
+  ADD COLUMN marked_for_deletion boolean; 
+  
+ALTER TABLE dbo.sccv_personal_relationship
+  ADD COLUMN marked_for_deletion boolean;   
+  
+CREATE TABLE IF NOT EXISTS dbo.sccv_person_record_to_be_deleted (
+    id bigint GENERATED BY DEFAULT AS IDENTITY NOT NULL CONSTRAINT xpk_person_record_to_be_deleted PRIMARY KEY,
+    fk_id_to_delete bigint,
+    fk_master_id bigint,
```
chubberlisk, line 529: What do you think about just naming these to be more explicit like `fk_person_id_to_delete` and `fk_master_person_id`?

database/schema.sql
```diff
@@ -500,3 +500,44 @@ update dbo.sccv_personal_relationship_type set inverse_type_id = (select id from
...
+        REFERENCES dbo.dm_persons(person_id),
+    CONSTRAINT fk_master_record
+        FOREIGN KEY(fk_master_id)
+        REFERENCES dbo.dm_persons(person_id)
+);
+
+CREATE TABLE IF NOT EXISTS dbo.sccv_deleted_person_record (
+    id bigint GENERATED BY DEFAULT AS IDENTITY NOT NULL CONSTRAINT xpk_deleted_person_record PRIMARY KEY,
+    deleted_id bigint,
+    master_id bigint,
```
chubberlisk, line 541: Same here, `deleted_person_id` and `master_person_id`? I know the table name tells you but I think it helps to be explicit in the column names as well to reduce cognitive load.

database/schema.sql
```diff
@@ -500,3 +500,44 @@ update dbo.sccv_personal_relationship_type set inverse_type_id = (select id from
...
+  ADD COLUMN marked_for_deletion boolean; 
+  
+ALTER TABLE dbo.sccv_allocations_combined
+  ADD COLUMN marked_for_deletion boolean; 
+  
+ALTER TABLE dbo.sccv_warning_note
+  ADD COLUMN marked_for_deletion boolean; 
+  
+ALTER TABLE dbo.sccv_personal_relationship
+  ADD COLUMN marked_for_deletion boolean;
```
chubberlisk, line 524: Just thinking do we also need to add this to the `sccv_personal_relationship_detail`? 🤔

chubberlisk (commented):
Thank you for writing such a detailed PR description as well as adding a test for the script! I've added some comments and suggestions below.

SocialCareCaseViewerApi.Tests/V1/Data/DeletePersonRecordsTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            var inversePersonalRelationship = new Faker<PersonalRelationship>()
+              .RuleFor(pr => pr.Id, f => f.UniqueIndex)
+              .RuleFor(pr => pr.PersonId, newMasterPersonRecord.Id)
+              .RuleFor(pr => pr.OtherPersonId, personToBeDeleted.Id)
+              .RuleFor(pr => pr.OtherPerson, personToBeDeleted)
+              .RuleFor(pr => pr.TypeId, personalRelationship.TypeId)
+              .RuleFor(pr => pr.StartDate, f => f.Date.Past())
+              .RuleFor(pr => pr.EndDate, f => f.Date.Future())
+              .RuleFor(pr => pr.IsInformalCarer, f => f.Random.String2(1, "YN"))
+              .RuleFor(pr => pr.ParentalResponsibility, f => f.Random.String2(1, "YN"));
```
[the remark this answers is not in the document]
LBHTKarki: Yes, that's a lot tidier. I have updated it. I ignored the relationship type ids for this as they are not really cricital for the test.

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
```diff
@@ -129,5 +129,29 @@ public void WhenThereAreMultipleRelationshipsReturnsAllRelevantPersonalRelations
...
+
+            DatabaseContext.Persons.Add(anotherPerson);
+            DatabaseContext.Persons.Add(personMarkedForDeletion);
+            DatabaseContext.PersonalRelationships.Add(anotherRelationship);
+            DatabaseContext.PersonalRelationships.Add(andAnotherRelationship);
+            DatabaseContext.SaveChanges();
+
+            var response = _databaseGateway.GetPersonWithPersonalRelationshipsByPersonId(person.Id);
+
+            response.PersonalRelationships.Should().HaveCount(2);
```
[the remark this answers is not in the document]
LBHTKarki: The `PersonalRelationshipsHelper.SavePersonWithPersonalRelationshipToDatabase` helper creates a person, second person and a relationship between them. Test then adds the third person and relationship for them. It then adds third relationship for the deleted person so the expected result is two.

database/schema.sql
LBHTKarki: that's better. Updated

database/schema.sql
LBHTKarki: I agree, updated

database/schema.sql
LBHTKarki: These flags are only added to entities directly linked to Person entity. For example I haven't included warning note reviews. Those will have to be included in the hard delete script, but I thought they are not necessary for the initial soft delete. Hard delete script will scan through the entity relationships and delete them all.

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/GetPersonWithPersonalRelationshipsByPersonIdTests.cs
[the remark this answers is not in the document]
chubberlisk: Ahhh okay, got it!

chubberlisk (approved):

PR #362, opened 2021-07-09
SCT-789 fix search by postcode

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-789

When post code is used as a filter in person search the current implementation includes records with historical (non display) addresses as well. This is confusing to users because the type, display vs non display, is not visible on the front end.

This updates the gateway to exclude non display matches from results so only current display addresses are returned. 

#### _Checklist_


JohnFarrellDev (approved):
Looks good

PR #364, opened 2021-07-13
SCT-641 enable search by emergency ids

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-641

Currently users have to prefix emergency IDs in person search. This is often problematic because users might not know what type of ID they have i.e. emergency or Mosaic.

Enable users to search person records by giving either emergency id or Mosaic id. Users don't have to provide prefix the emergency ids anymore.

This also removes the logic the were only emergency IDs with NC prefix were supported. This is important so we can support them all in the future.

New implementation will automatically find all matching person records regardless of what id type was used. 

#### _Checklist_

Inform users that it is no longer necessary to use the prefix although they can still provide it if they want.


SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        [Test]
+        public void IfLimitMoreThanTheMaximumWillUseTheMaximumLimit()
+        {
+            _mockDatabaseGateway.Setup(x => x.GetResidentsBySearchCriteria(0, 100, null, null, null, null, null, null, null))
+                .Returns(new List<ResidentInformation>()).Verifiable();
+
+            _getAllUseCase.Execute(new ResidentQueryParam(), 0, 400);
+
+            _mockDatabaseGateway.Verify();
+        }
```
chubberlisk, line 73: I think we could improve the readability and understanding of these tests as I struggled at first. For example, seeing the lone `_mockDatabaseGateway.Verify();` as the expectation was initially confusing. I eventually got it and understood that we're expecting that the database gateway to be called with a limit of 10/100 but because we've stubbed the gateway method to only reply to those arguments so we can just use `Verify()`.

Maybe we can remove the stubbing here as we're already doing it in the `SetUp` and update the expectation `Verify()` to check the arguments to be 10/100?

I also think it would help to extract variables for these numbers e.g. `var maximumLimit = 100` so it's clearer.

chubberlisk: Possibly silly question but what happens on the boundary when the limit is 10? 😅

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        public void ReturnsTheNextCursor()
+        {
+            var residents = _fixture.CreateMany<ResidentInformation>(10);
+            int idCount = 10;
+            foreach (ResidentInformation resident in residents)
+            {
+                idCount++;
+                resident.MosaicId = idCount.ToString();
+            }
+            var expectedNextCursor = residents.Max(r => r.MosaicId);
```
chubberlisk, line 85: Could you explain what we're doing here in the setup? Why do we need to change the `MosaicId`s of the residents?

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            var listOfMatchingIds = new List<long> { 1, 2 };
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonIdsByEmergencyId(It.IsAny<long>())).Returns(listOfMatchingIds);
+            _mockDatabaseGateway.Setup(x => x.GetPersonsByListOfIds(It.IsAny<List<long>>())).Returns(new List<Person>() { TestHelpers.CreatePerson() });
+
+            _getAllUseCase.Execute(request, cursor: 0, limit: 4);
+
+            _mockDatabaseGateway.Verify(x => x.GetPersonsByListOfIds(listOfMatchingIds));
+        }
```
chubberlisk, line 183: What do you think about adding `TestCase`s for different `MosaicId`s? As the code mentions we want to be removing prefixes and other letters.

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
LBHTKarki: I have updated the setup and used constants for the limits. Hopefully it's more clear now. I've also added tests for the boundary values to make it clear what happens in those situations.

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
LBHTKarki: When fixture library creates those person objects it uses Ids like MosaicId18e8cb3e-d6d8-4e4b-9165-3ad8413d38a5. That idCount together with the loop resets the IDs to values we use (long), so the next cursor calculation can be performed correctly and reflects the types we use.

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
LBHTKarki: I've added test cases for each

SocialCareCaseViewerApi.Tests/V1/UseCase/Residents/GetAllUseCaseTests.cs
chubberlisk: Ahhhhh okay, I see thanks!

chubberlisk (approved):
Thanks for addressing my comments! 🙇‍♀️

PR #366, opened 2021-07-13
Add default value for marked for deletion flag

LBHTKarki:

Marked for deletion flag doesn't have default value which means all records get filtered out by APIs.

Set default value to false so the API returns correct records.

#### _Checklist_

Run the alter table scripts in the PR


chubberlisk (commented):
Will setting it default mean all current values will be `false`? By that I mean do we have to set as default and then run an update statement to set current ones to `false`?

LBHTKarki:
> Will setting it default mean all current values will be `false`? By that I mean do we have to set as default and then run an update statement to set current ones to `false`?

@chubberlisk You are correct, update statements are also required. I have pushed an update to add those.

LBHTKarki:
@chubberlisk I've pushed the manual update plan to this same PR for simplicity

chubberlisk (approved):
👍

PR #369, opened 2021-07-14
SCT-833 - Add a new temporary object to the BE response

mapleleafjack:
## Link to JIRA ticket

[Link to ticket](https://hackney.atlassian.net/browse/SCT-833)

## Describe this PR

A "relationship" object has been added to the Personal Relationship response.
This object is very similar to the pre-existing "persons" object, with the difference that it includes the RelationshipID (as "id") together with the PersonID (as "personId").

The RelationshipID is necessary for the relationship deletion request from the Frontend.

Added a second object "relationship" in the Relationship response.

#### _Checklist_

This can be removed after the BE and FE are aligned


SocialCareCaseViewerApi/V1/Domain/RelatedRelationship.cs
```diff
@@ -0,0 +1,14 @@
+namespace SocialCareCaseViewerApi.V1.Domain
+{
+    public class RelatedRelationship
```
chubberlisk, line 3: I'm sure we could name this better but I can't think of anything! 😅

chubberlisk (approved):
Looks good to me!

SocialCareCaseViewerApi/V1/Domain/RelatedRelationship.cs
mapleleafjack: Maybe "Relationship" or "Relation" can be used, anyways we can change it when we remove the code for "RelatedPerson"

chubberlisk: Hmmm, not sure about those as they feel too generic and possibly confusing against other classes. Let's run with this for now and like you said, we can revisit the name.

SocialCareCaseViewerApi/V1/Domain/RelatedRelationship.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public long Id { get; set; }
+        public long PersonId { get; set; }
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public string Gender { get; set; }
+        public string IsMainCarer { get; set; }
+        public string IsInformalCarer { get; set; }
+        public string Details { get; set; }
```
JohnFarrellDev, line 12: Details can be null right?

Can any of the other properties on RelatedRelationship.

I'd advise being explicit about what can and can't be null (https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references)

mapleleafjack: Yes, Details can be null!

I'll set it nullable and add the #nullable enable annotation right on top of it

JohnFarrellDev (approved):
:rocket:

PR #370, opened 2021-07-14
fix(SCT-471): fix not auditing personal relationship and details

chubberlisk:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-471

We're not correctly auditing `PersonalRelationship` and `PersonalRelationshipDetails` because we're missing the `CreatedBy` attribute and `IAuditEntity` for `PersonalRelationshipDetails`.

The frontend has been updated to send the user's email as `createdBy` in the request, see https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/567.

This PR updates the `CreatePersonalRelationshipRequest` to take in `CreatedBy` and then ensures we save this for a `PersonalRelationship` and `PersonalRelationshipDetails`. It also adds the missing `IAuditEntity` for `PersonalRelationshipDetails`.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
@@ -136,5 +136,37 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways.Database
             details?.PersonalRelationshipId.Should().Be(response.Id);
             details?.Details.Should().Be(request.Details);
         }
+
+        [Test]
+        public void AuditsThePersonalRelationship()
+        {
+            var (person, otherPerson) = PersonalRelationshipsHelper.SavePersonAndOtherPersonToDatabase(DatabaseContext);
+            var type = DatabaseContext.PersonalRelationshipTypes.FirstOrDefault(prt => prt.Description == "parent");
+            var request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(
+                person.Id, otherPerson.Id, type.Id, type.Description
+            );
+
+            var response = _databaseGateway.CreatePersonalRelationship(request);
+
+            var personalRelationship = DatabaseContext.PersonalRelationships.FirstOrDefault();
+            personalRelationship?.CreatedAt.Should().NotBeNull();
+            personalRelationship?.CreatedBy.Should().Be(request.CreatedBy);
+        }
+
+        [Test]
+        public void AuditsThePersonalRelationshipDetails()
+        {
+            var (person, otherPerson) = PersonalRelationshipsHelper.SavePersonAndOtherPersonToDatabase(DatabaseContext);
+            var type = DatabaseContext.PersonalRelationshipTypes.FirstOrDefault(prt => prt.Description == "parent");
+            var request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(
+                person.Id, otherPerson.Id, type.Id, type.Description
+            );
+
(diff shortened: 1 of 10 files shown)


mapleleafjack (approved):

SocialCareCaseViewerApi/V1/Boundary/Requests/CreatePersonalRelationshipRequest.cs
```diff
@@ -45,6 +48,8 @@ public CreatePersonalRelationshipRequestValidator()
                 .Matches("(?i:^Y|N)$").WithMessage("'isInformalCarer' must be 'Y' or 'N'.");
             RuleFor(pr => pr.Details)
                 .MaximumLength(1000).WithMessage("'details' must be less than or equal to 1,000 characters.");
+            RuleFor(pr => pr.CreatedBy)
+                .NotNull().WithMessage("'createdBy' must be provided.");
```
JohnFarrellDev, line 52: Should we also be confirming that createdBy is an email

JohnFarrellDev: It would probably also be a good idea in the usecase PersonalRelationshipsUseCase#ExecutePost to confirm the createdBy email is a worker in the system, otherwise throw `WorkerNotFoundException` or something you think is sensible

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreatePersonalRelationshipTests.cs
```diff
@@ -136,5 +136,37 @@ public void CreatesAPersonalRelationshipDetails()
...
+            var (person, otherPerson) = PersonalRelationshipsHelper.SavePersonAndOtherPersonToDatabase(DatabaseContext);
+            var type = DatabaseContext.PersonalRelationshipTypes.FirstOrDefault(prt => prt.Description == "parent");
+            var request = PersonalRelationshipsHelper.CreatePersonalRelationshipRequest(
+                person.Id, otherPerson.Id, type.Id, type.Description
+            );
+
+            var response = _databaseGateway.CreatePersonalRelationship(request);
+
+            var personalRelationship = DatabaseContext.PersonalRelationships.FirstOrDefault();
+            personalRelationship?.CreatedAt.Should().NotBeNull();
```
JohnFarrellDev, line 152: I'd rather our test not just confirm the value isn't null.
However this is a bigger piece of work so I'll create a ticket.

I expect the issue is that CreatedAt is happening at runtime during the actual object creation so we have no way of validating the actual value against something we expected. The datetime will be new every time the test runs. 
We need something to abstract DateTime which we can then mock for our tests.

https://stackoverflow.com/questions/2425721/unit-testing-datetime-now

JohnFarrellDev (approved):
:+1:

SocialCareCaseViewerApi/V1/Boundary/Requests/CreatePersonalRelationshipRequest.cs
chubberlisk: Ahhh these are very good shouts, thanks!

chubberlisk: Updated so the use case checks and the API endpoint responds with 400.

chubberlisk: And updated so we check the `createdBy` in the request is an email address! (Totally didn't nearly forget about that! Cool we can do this so easily with FluentValidation. 😎)

PR #378, opened 2021-07-19
Sct 867 simple audit feature

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-896

This is related to to the feat(SCT-895) where we need to add some basic tracking details for get `residents/{id}` and `/cases/{id}` end points.

Audit details will be captured based on the `auditingEnabled` query string parameter. The process must be non-blocking i.e. we can't have validation on the incoming audit paramater values and also we can't throw an exception should something go wrong when the record is added to the database. In order to have some visibility on possible errors, I'll look into lambda logger option on a separate PR.  

This is not the proposed solution for generic access logging, but just a minimum/temporary setup so we can capture some details for those two end points. 

Get case by id boundary object had to be tweaked a bit to support the capture of those audit values. It is now using an object rather than simple string. This will be reverted back after the audit requirements in this format are no longer required. Also this update is backwards compatible with the current implementation, so the end point can be called with simple id parameter of with full set of audit parameters.

Logic for the audit requirement is in the controller and also the request object for the use case is constructed in the controller to keep the use case generic.

#### _Checklist_

Apply schema changes to database and test the full setup on staging


SocialCareCaseViewerApi.Tests/V1/Controllers/CaseControllerTests.cs
```diff
@@ -83,5 +92,64 @@ public void ListCasesReturns404WhenNoCasesAreFound()
             response?.StatusCode.Should().Be(404);
             response?.Value.Should().Be("Document Not Found");
         }
+
+        [Test]
+        public void GetCaseByRecordIdDoesNotCallTheCreateRequestAuditUseCaseWhenAuditingIsEnabledIsFalse()
+        {
+            var request = new GetCaseNotesRequest();
```
chubberlisk, line 99:
```suggestion
            var request = new GetCaseNotesRequest()
             {
                 AuditingEnabled = false,
                 UserId = _faker.Person.Email,
                 Id = "tyut67t89t876t",
                 ResidentId = "4"
             };
```

I think this is more reflective of the test description we have. It also might be worth adding a test which just has `Id` to prove we're backwards compatible.

SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
```diff
@@ -114,6 +116,49 @@ public void GetPersonByIdReturns200WhenSuccessful()
             response?.StatusCode.Should().Be(200);
         }
 
+        [Test]
+        public void GetPersonByIdDoesNotCallTheCreateRequestAuditUseCaseWhenAuditingIsEnabledIsFalse()
+        {
+            var request = new GetPersonRequest();
```
chubberlisk, line 122: Similar thought as above.

chubberlisk (approved):

database/schema.sql
```diff
@@ -1,587 +1,595 @@
...
+    timestamp timestamp NOT NULL DEFAULT now()
+);
+
+CREATE TABLE IF NOT EXISTS dbo.sccv_request_audit(
+   id bigint GENERATED BY DEFAULT AS IDENTITY NOT NULL CONSTRAINT xpk_sccv_request_audit PRIMARY KEY,
+   timestamp timestamp NOT NULL,
+   action_name varchar(200) NOT NULL,				
+   user_name varchar(200) NOT NULL,			
+   metadata json NOT NULL
+);
```
chubberlisk, line 595: Don't forget to add a manual update so we keep track of what we're doing! 🕵️‍♀️

SocialCareCaseViewerApi.Tests/V1/Controllers/SocialCareCaseViewerApiControllerTests.cs
LBHTKarki: Good point, I've updated the tests and added new ones for backwards compatibility check as well.

PR #388, opened 2021-07-26
chore: Update the make restart db command

Renny4Real:
The make command to restart the test databases was referencing the old image name of the Postgres test database.

We use `make restart-db` to make sure our local docker databases are up to date with any schema changes in the current branch.

This updates the command to reference the new image and service name for the docker test database.

This might not work if the old `test-database` image still exists
or if the new image name: `sccv-api-test-postgresql` does not exist.

Would recommend checking and cleaning up the docker images to match what we be expecting before running this command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,10 @@ start-test-dbs:
 
 .PHONY: restart-db
 restart-db:
-	docker stop $$(docker ps -q --filter ancestor=test-database -a)
-	-docker rm $$(docker ps -q --filter ancestor=test-database -a)
-	docker rmi test-database
-	docker-compose up -d test-database
+	docker stop $$(docker ps -q --filter ancestor=sccv-api-test-postgresql -a)
+	-docker rm $$(docker ps -q --filter ancestor=sccv-api-test-postgresql -a)
+	docker rmi sccv-api-test-postgresql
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,10 +28,10 @@ start-test-dbs:
 
 .PHONY: restart-db
 restart-db:
```
chubberlisk, line 30: Should we also restart the MongoDB?

If so, then I think we should change the command name to be `restart-test-dbs` so it's clear it's both databases and it's for the test ones.

If not, then I think we should change the command name to be `restart-test-postgres-db` or `restart-test-pg-db` to be clear it's just for the PostgreSQL database and it's for the test ones.

chubberlisk (approved):
Tested the command and worked for me! Just one comment/suggestion.

Makefile
Renny4Real: Not sure if we need to restart the mongo db 🤔.

I haven't had any issues related to out of sync mongo DBs.

We could rename this to be clearer that this would affect only the Postgres DB and maybe if we have issues with inconsistent mongo DBs, we add a command for it later?

chubberlisk: Coooool, let's go with what you suggested! 🚢

PR #389, opened 2021-07-27
SCT-815 update mongodb record with new mosaic id

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-815

When person records are soft deleted in postgreSQL the existing form records must be transferred over to the new master person id. This scripts updates all relevant records for a given ID in `form_data` collection.

Comments from the script:
**summary**
 This script is designed to be run against a single ID as a workaround until the appropriate API endpoint is in place
 Only runs against form_data collection
 Updates all matching records that are not any of the audit event types created by the service API

**Howto**
To run this script manually on the server:
- remove double quotes
- add appropriate IDs (one in the query (q) is the current one and the one in the $set is the new master id)
- pass it to the db.runCommand() like this:

```
db.runCommand({
 update: "form_data",
 updates:
     [
         {
              q: { MosaicId: "334445566", FormNameOverall: { $not: /^API_/ } }, u: { $set: { MosaicId: "77889900" } }, multi: true
         }
     ]
});
```

Added a script that can be run against form_data collection to update the form records.
Please note this does not handle any other collections, so residents submissions collection for example has to be handled separately (done in separate PR once the transfer method is agreed).

Re-structured data script tests to new namsespaces

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
@@ -104,9 +104,9 @@ namespace SocialCareCaseViewerApi.Tests.V1.Data.MongoDB
             string mosaicIdTwo = "556677";
             string masterPersonId = "77889900";
 
-            var casenoteOne = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
-            var casenoteTwo = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
-            var casenoteThree = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicIdTwo).Create();
+            var casenoteOneForMosaicId = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
+            var casenoteTwoForMosaicId = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
+            var casenoteThreeForMosaicIdTwo = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicIdTwo).Create();
 
             var collection = _mongoDatabase?.GetCollection<MongoDBTestObject>(MainFormCollection);
 
--- a/SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
@@ -6,6 +6,7 @@ using MongoDB.Driver;
 using NUnit.Framework;
 using SocialCareCaseViewerApi.V1.Infrastructure;
 using System;
+using System.Linq;
 
 #nullable enable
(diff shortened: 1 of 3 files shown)


SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
```diff
@@ -0,0 +1,156 @@
...
+        [Test]
+        public void DoesNotUpdateOtherThanMatchingRecords()
+        {
+            string mosaicId = "334445566";
+            string mosaicIdTwo = "556677";
+            string masterPersonId = "77889900";
+
+            var casenoteOne = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
+            var casenoteTwo = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
+            var casenoteThree = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicIdTwo).Create();
```
chubberlisk, line 109:
```suggestion
            var casenoteOneForMosaicId = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
            var casenoteTwoForMosaicId = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicId).Create();
            var casenoteThreeForMosaicIdTwo = _fixture.Build<MongoDBTestObject>().With(x => x.MosaicId, mosaicIdTwo).Create();
```

Or something like that to make it clear that the third case note is for another person.

SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            collection?.InsertOne(casenoteOne);
+            collection?.InsertOne(casenoteTwo);
+            collection?.InsertOne(casenoteThree);
+
+            var result = _mongoDatabase?.RunCommand(_dbCommandUnderTest);
+
+            var filterByNewId = Builders<MongoDBTestObject>.Filter.Eq("MosaicId", masterPersonId);
+            var recordsWithNewId = collection.Find(filterByNewId).ToList();
+            recordsWithNewId.Count.Should().Be(2);
```
chubberlisk, line 148: What do you think of changing the expectation to find the case note with the `FormNameOverall` is `API_Event` and checking that the `MosaicId` is still `mosaicId`? I think that's more precise than checking the count.

chubberlisk (approved):

SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
LBHTKarki: @chubberlisk I've renamed them. That's better

SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/UpdateMosaicIdFormDataCollection.cs
LBHTKarki: @chubberlisk That's more robust, I've pushed an update.

PR #391, opened 2021-07-28
Fix(SCT-270): Updating a worker team should remove associations with their old teams

Renny4Real:
## Link to JIRA ticket
https://hackney.atlassian.net/browse/SCT-270

From a user’s perspective, the issue is that when a user’s team is updated, their associated allocations are expected to have the updated team name and not the previous team. So after updating a worker, the allocations for the worker are still referencing the old team.

i.e. Users are expecting any update of a worker’s team to reflect in the allocations associated with the worker.

However, when the worker's team does look like its updated in the UI. The team is added to the worker, so now the worker is assigned to multiple teams but the UI would grab just the first one returned in the list.

This adds an integration test to replicate the user journey of updating a worker team and expecting any worker info page and any allocations for the person to be updated with the updated team as well. 

When a worker's team is updated, all previous associated teams in the `workerteam` table are removed and only the team in the update request is added as the only team for the worker. (Previously, this would just add the new team to a list of teams for the worker but now removes all previous associations and adds the new one)

When a worker's team is updated, any allocations already assigned to that worker prior to the update, have their details updated to use the new team. (Previously, this had the previous team set on the allocation object but now, when a worker's team is updated, we update every allocation to use the new team and save the changes in the database)

This also adds new setup that should hopefully help with adding future integration tests.

#### _Checklist_

- [X] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature

In Staging, test that when a worker's team is updated: 
- The worker's information page displays the new team,
- The response from getting a worker, only has one team in it i.e the updated team
- Any allocations that existed prior to the change are also updated once the update occurs.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -274,43 +274,54 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         [Test]
         public void UpdateWorkerUpdatesTheTeamSetOnAnyAllocations()
         {
+            // Create worker and teams
             var worker = TestHelpers.CreateWorker(hasAllocations: false, hasWorkerTeams: false, id: 123);
+            var differentTeam = TestHelpers.CreateTeam(name: "different team X", context: worker.ContextFlag);
             var workerTeam = TestHelpers.CreateWorkerTeam(worker.Id);
             worker.WorkerTeams = new List<WorkerTeam> { workerTeam };
 
             DatabaseContext.Workers.Add(worker);
             DatabaseContext.WorkerTeams.Add(workerTeam);
-
-            var differentTeam = TestHelpers.CreateTeam(teamId: 2000, name: "different team X", context: worker.ContextFlag);
             DatabaseContext.Teams.Add(differentTeam);
-
             DatabaseContext.SaveChanges();
 
-            var (createAllocationRequest, _, createdByWorker, person, _) = TestHelpers.CreateAllocationRequest(workerId: worker.Id, teamId: workerTeam.TeamId);
-            DatabaseContext.Workers.Add(createdByWorker);
-            DatabaseContext.Persons.Add(person);
-            DatabaseContext.SaveChanges();
-
-            _classUnderTest.CreateAllocation(createAllocationRequest);
-
+            // Check worker's details before updating
             var originalWorker = _classUnderTest.GetWorkerByWorkerId(worker.Id);
-            var originalTeam = DatabaseContext.WorkerTeams.Single(x => x.WorkerId == originalWorker.Id).Team;
+            var currentTeam = DatabaseContext.WorkerTeams.Single(x => x.WorkerId == originalWorker.Id).Team;
 
             originalWorker.WorkerTeams?.Count.Should().Be(1);
+            originalWorker.WorkerTeams?.Single().TeamId.Should().Be(currentTeam.Id);
+            originalWorker.Allocations?.Count.Should().Be(0);
+
+            var (createAllocationRequest, _, allocator, resident, _) = TestHelpers.CreateAllocationRequest(workerId: worker.Id, teamId: workerTeam.TeamId);
+            DatabaseContext.Workers.Add(allocator);
+            DatabaseContext.Persons.Add(resident);
+            DatabaseContext.SaveChanges();
+
+            // Check allocations assigned to the worker
+            _classUnderTest.CreateAllocation(createAllocationRequest);
             originalWorker.Allocations?.Count.Should().Be(1);
 
-            var originalAllocation = _classUnderTest.SelectAllocations(0, workerId: originalWorker.Id);
+            var allocation = _classUnderTest.SelectAllocations(0, workerId: originalWorker.Id);
 
-            originalAllocation.Count.Should().Be(1);
-            originalAllocation.Single().AllocatedWorkerTeam.Should().Be(originalTeam.Name);
+            allocation.Count.Should().Be(1);
+            allocation.Single().AllocatedWorkerTeam.Should().Be(currentTeam.Name);
 
-            var updateRequest = TestHelpers.CreateUpdateWorkersRequest(teamId: differentTeam.Id, teamName: differentTeam.Name,
-                workerId: worker.Id, firstName: worker.FirstName, lastName: worker.LastName, role: worker.Role, contextFlag: worker.ContextFlag);
+            // Update worker to be in a new team
+            var updateTeamRequest = TestHelpers.CreateUpdateWorkersRequest(teamId: differentTeam.Id, teamName: differentTeam.Name,
+                workerId: originalWorker.Id, firstName: originalWorker.FirstName,
+                lastName: originalWorker.LastName, role: originalWorker.Role, contextFlag: originalWorker.ContextFlag);
 
-            _classUnderTest.UpdateWorker(updateRequest);
+            _classUnderTest.UpdateWorker(updateTeamRequest);
 
-            var getUpdatedWorker = _classUnderTest.GetWorkerByWorkerId(worker.Id);
+            // Check worker's details after updating
+            var getUpdatedWorker = _classUnderTest.GetWorkerByWorkerId(originalWorker.Id);
 
+            getUpdatedWorker.WorkerTeams?.Count.Should().Be(1);
+            getUpdatedWorker.WorkerTeams?.Single().TeamId.Should().Be(differentTeam.Id);
+            getUpdatedWorker.Allocations?.Count.Should().Be(1);
+
+            // Check allocations assigned to the worker have been updated
             var updatedAllocations = _classUnderTest.SelectAllocations(0, workerId: getUpdatedWorker.Id);
             updatedAllocations.Count.Should().Be(1);
             updatedAllocations.Single().AllocatedWorkerTeam.Should().Be(differentTeam.Name);
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using AutoFixture;
 using Bogus;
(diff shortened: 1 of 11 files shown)


SocialCareCaseViewerApi.Tests/V1/IntegrationTests/UpdateWorkerDetails.cs
```diff
@@ -0,0 +1,68 @@
...
+            // Get request to check team has been updated
+
+            var getUpdatedWorkersResponse = await Client.GetAsync(getWorkersUri).ConfigureAwait(true);
+            getUpdatedWorkersResponse.StatusCode.Should().Be(200);
+
+            var updatedContent = await getUpdatedWorkersResponse.Content.ReadAsStringAsync().ConfigureAwait(true);
+            var updatedWorkerResponse = JsonConvert.DeserializeObject<List<WorkerResponse>>(updatedContent).ToList();
+
+            updatedWorkerResponse.Count.Should().Be(1);
+            updatedWorkerResponse.Single().Teams.Count.Should().Be(1);
```
Renny4Real, line 63: I was expecting this line to fail and error with a message related to 2 teams being found.
When you look at the response when we update a worker and view their details in the UI, the response from the API has a more than one element in the teams list for the worker.

Not sure why I can't recreate the same process here 🤔 . Might have missed something?

Renny4Real: Have been able to recreate this in 03ff05b

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -272,6 +271,51 @@ public void UpdateWorkerUpdatesAnExistingWorkerInDatabase()
...
+
+            originalAllocation.Count.Should().Be(1);
+            originalAllocation.Single().AllocatedWorkerTeam.Should().Be(originalTeam.Name);
+
+            var updateRequest = TestHelpers.CreateUpdateWorkersRequest(teamId: differentTeam.Id, teamName: differentTeam.Name,
+                workerId: worker.Id, firstName: worker.FirstName, lastName: worker.LastName, role: worker.Role, contextFlag: worker.ContextFlag);
+
+            _classUnderTest.UpdateWorker(updateRequest);
+
+            var getUpdatedWorker = _classUnderTest.GetWorkerByWorkerId(worker.Id);
```
JohnFarrellDev, line 312: Should we have a test against `getUpdatedWorker` and also `originalWorker`, show that `originalWorker.WorkerTeams` is a length of 1 and that the teamId matches the `workerTeam`. Then test that the `updatedWorker` still has `WorkerTeams` of length 1 but the teamId now matches `differentTeam`.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestHelpers.cs
```diff
@@ -0,0 +1,175 @@
...
+                .RuleFor(p => p.PersonIdLegacy, f => f.Random.String2(16))
+                .RuleFor(p => p.AgeContext, f => ageContext ?? f.Random.String2(1, "AC"))
+                .RuleFor(p => p.DataIsFromDmPersonsBackup, f => f.Random.String2(1, "YN"))
+                .RuleFor(p => p.SexualOrientation, f => f.Random.String2(10))
+                .RuleFor(p => p.PreferredMethodOfContact, f => f.Random.String2(10))
+                .RuleFor(p => p.Restricted, f => restricted ?? f.Random.String2(1, "YN")).Generate();
+
+            var insertPersonQuery = SeedPerson(person);
+
+            context.Database.ExecuteSqlRaw(insertPersonQuery);
```
JohnFarrellDev, line 39: Is there a reason we have to use these raw SQL strings to insert a person into the DB? Could we not use the DatabaseGateWay and take advantage of EF? I might be missing something but this seems like a more complex way of doing this. If our Person changes then we need to update our SQL for the tests.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestHelpers.cs
```diff
@@ -0,0 +1,175 @@
...
+                .RuleFor(w => w.CreatedAt, f => f.Date.Soon())
+                .RuleFor(w => w.LastModifiedAt, f => f.Date.Soon())
+                .RuleFor(w => w.IsActive, true)
+                .RuleFor(w => w.Allocations, new List<AllocationSet>())
+                .RuleFor(w => w.WorkerTeams, new List<WorkerTeam> { workerTeam })
+                .RuleFor(w => w.LastModifiedBy, f => f.Person.Email).Generate();
+
+            var insertWorkerQuery = SeedWorker(worker);
+
+            context.Database.ExecuteSqlRaw(insertWorkerQuery);
```
JohnFarrellDev, line 87: Same question as above.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestSetup.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        private MockWebApplicationFactory<TStartup> _factory;
+        private NpgsqlConnection _connection;
+        private NpgsqlTransaction _transaction;
+        private DbContextOptionsBuilder _builder;
+
+
+        protected HttpClient Client => _client;
+        protected DatabaseContext DatabaseContext => _databaseContext;
+        protected ISccvDbContext MongoDbTestContext => _mongoDbContext;
```
JohnFarrellDev, line 30: Can delete this line

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestSetup.cs
```diff
@@ -0,0 +1,91 @@
+using System;
+using System.Data;
+using System.Data.Common;
+using System.Net.Http;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage;
```
JohnFarrellDev, line 6: Couple of unused imports

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/UpdateWorkerDetails.cs
```diff
@@ -0,0 +1,140 @@
...
+            _existingDbWorker = existingDbWorker;
+            _existingDbTeam = existingDbTeam;
+            _allocationWorker = allocationWorker;
+            _differentDbTeam = differentDbTeam;
+            _resident = resident;
+        }
+
+
+        [Test]
+        public async Task UpdateWorkerWithNewTeamReturnsTheOnlyTheUpdatedTeam()
```
JohnFarrellDev, line 55: If we're testing the updating of a worker I think it can be useful to have assertions before the patch. To me it proves more strongly that the PATCH request did change the team the worker belongs to.

JohnFarrellDev:
I think this work is really good and important for the project. Because some of the integration test stuff is fairly complex would you be able to spend some time going through parts of this PR with me?

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
Renny4Real: We could 🤔 
We did this previously but the test was returning a false positive. I think I remember reading that entity framework would remember the changes on an entity if it had touched it once or something like that. This false positive is what inspired having an integration test but does make sense that we can also assert at the gateway level, that what we want happens here too.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestHelpers.cs
Renny4Real: The reason for using the `ExecuteSqlRaw()` was so that we don't have the query triggering a transaction.

One of the issues was getting the test database populated with the data. After asking around a bit, it was suggested that the reason the data wasn't getting to our database was because we were using transactions and for those changes to save to the database we needed to commit the transaction and it turns out when we call `Database.SaveChanges()` in our actual code, it doesn't commit the transaction the integration test. 🤷🏾‍♀️ 

Adding `transaction.Commit()` saved the data to the DB but it didn't replicate the bug when running the test. 

It's still unclear to me but happy to dig into this a bit better later but it was returning a false positive as well and would replace the `workerteam` data in the test database rather than add to it which is what we're trying to replicate it. 

Based on all the stuff we'd researched to understand what might be going on, my hypothesis was that if we seeded the test database in way that entity framework doesn't know about the data initially, we could then run the rest of the code using entity framework, and this replicated the bug.

The hypothesis stems from the fact that we'd figured that there was something different in the setup of our test database compared to the local dev database since our tests followed the user journey properly. My current understanding is the possibility of some magic happening when Entity Framework is tracking entities that is not entirely reflected in the actual database.

But happy to try and dig into this a bit further.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/UpdateWorkerDetails.cs
Renny4Real: I agree (and I can try this again) but it came back to entity framework already knowing about the worker before it changed and led to the false positive. I'm not sure if it's the test setup or if there are ways to set entity framework to not do some of this hard to understand magic 😅 but yeah, I think I tried getting the worker first and checking their team but then entity framework did its magic because now it knew and was tracking it (which is my best guess for why the test returned a false positive when you get the worker first. 😅 

But happy to check this again

Renny4Real:
@JohnFarrellDev Yeah definitely, I think taking someone through would be helpful. There might be things here from investigating that might not be needed anymore or things that could be added and cleaned up a bit more. I definitely want to be able to add some documentation around this too so would be helpful to know what is useful to explain and add. 🙂

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestHelpers.cs
LBHTKarki: We've had issues with tracking in gateway tests before and have disabled it in certain tests. That might help in your use case as well. So when you initialise your DatabaseContext you can disable the tracker:  `DatabaseContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;`

Renny4Real: @LBHTKarki Yeah, Ting mentioned this and when we tried it the data didn't update. Checking the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontextoptionsbuilder.usequerytrackingbehavior?view=efcore-5.0) we found that we shouldn't disable tracking if we want to manipulate and persist changes.

LBHTKarki: @Renny4Real yes, disabling it for the actual implementation is not a good idea and we use it for automated audit for example. It just sometimes useful in tests when you are dealing with related entities. For example when you create related entities your tests might be passing even if your gateway queries are not setup correctly. You'll only notice the issue at runtime. Tracking combined with tansactions in tests can result in some challening situations like this one. Glad you are tackling integration tests and this setup is really useful for future tests

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
Renny4Real: Done in fe36ffb

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -272,6 +271,62 @@ public void UpdateWorkerUpdatesAnExistingWorkerInDatabase()
...
+                workerId: originalWorker.Id, firstName: originalWorker.FirstName,
+                lastName: originalWorker.LastName, role: originalWorker.Role, contextFlag: originalWorker.ContextFlag);
+
+            _classUnderTest.UpdateWorker(updateTeamRequest);
+
+            // Check worker's details after updating
+            var getUpdatedWorker = _classUnderTest.GetWorkerByWorkerId(originalWorker.Id);
+
+            getUpdatedWorker.WorkerTeams?.Count.Should().Be(1);
+            getUpdatedWorker.WorkerTeams?.Single().TeamId.Should().Be(differentTeam.Id);
```
Renny4Real, line 321: N.B: This assertions currently passes when it fails but added here to document what we would expect at the gateway level

Renny4Real:
Merging this to get the bug fixes in i.e When updating a worker's team:
* New team shows in the worker's details
* New team shows in any allocations they have

We can further iterate on the design of integration tests. 🎉

PR #392, opened 2021-07-29
feat(SCT-961): Capture date of event from answers

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-961)

Some of our forms ask the user for the date of event, we should be capturing this value.
This will allow us to store forms against a resident from the date of event instead of when the form was created.

Added a new property DateOfEvent to our CaseSubmission.
Capture DateOfEvent from controller.
In usecase if DateOfEvent provided with answers then update the CaseSubmission DateOfEvent property.
Updated CaseSubmission to ToCareCaseData to utilise the new DateOfEvent property if it exists.
Remove references to tags, this is a property our forms no longer use.
Tested.

#### _Checklist_

Do the frontend piece of work to make sure the DateOfEvent is sent in the request with the answers.

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
@@ -506,9 +506,8 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         public void CaseSubmissionToCareCaseDataReturnsCaseFormTimeStampAsDateTimeNowIfSubmittedAtNull()
         {
             var residents = new List<Person> { TestHelpers.CreatePerson() };
-            var workers = new List<dbWorker> { TestHelpers.CreateWorker(), TestHelpers.CreateWorker() };
             var request = TestHelpers.CreateListCasesRequest(residents[0].Id);
-            var submission = TestHelpers.CreateCaseSubmission(workers: workers, residents: residents, submittedAt: null);
+            var submission = TestHelpers.CreateCaseSubmission(residents: residents, submittedAt: null);
 
             var response = submission.ToCareCaseData(request);
 
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
@@ -499,7 +499,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
 
             var response = submission.ToCareCaseData(request);
 
-            response.CaseFormTimestamp.Should().Be(submittedAt.ToString("yyyy-MM-dd"));
+            response.CaseFormTimestamp.Should().Be("2021-07-20");
         }
 
         [Test]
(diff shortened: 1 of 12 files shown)


SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
```diff
@@ -378,7 +379,7 @@ public void CanMapPhoneNumberFromDomainToInfrastructure()
 
             var infrastructurePhoneNumber = new dbPhoneNumber()
             {
-                Number = phoneNumber.ToString(),
+                Number = phoneNumber,
```
JohnFarrellDev, line 382: Unrelated to the actual purpose of the PR but no need to call `.ToString()` against something that is already type `String`

SocialCareCaseViewerApi/V1/Controllers/FormSubmissionController.cs
```diff
@@ -78,11 +78,6 @@ public IActionResult CreateSubmission([FromBody] CreateCaseSubmissionRequest req
             {
                 var createdSubmission = _formSubmissionsUseCase.ExecutePost(request).Item1;
 
-                if (createdSubmission.SubmissionId == null)
-                {
-                    return StatusCode(500, "Case submission created with a null submission ID");
-                }
-
```
JohnFarrellDev, line 85: Deleted because SubmissionId could never be null, the if conditional could never be true

SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
```diff
@@ -460,6 +461,84 @@ public void CreateCaseNoteRequestToCaseNotesDocumentUsesCorrectDateTimeFormatFor
...
+        public void CaseSubmissionToCareCaseDataReturnsCaseFormTimeStampAsDateTimeNowIfSubmittedAtNull()
+        {
+            var residents = new List<Person> { TestHelpers.CreatePerson() };
+            var request = TestHelpers.CreateListCasesRequest(residents[0].Id);
+            var submission = TestHelpers.CreateCaseSubmission(residents: residents, submittedAt: null);
+
+            var response = submission.ToCareCaseData(request);
+
+            response.CaseFormTimestamp.Should().Be(DateTime.Now.ToString("yyyy-MM-dd"));
+        }
```
JohnFarrellDev, line 515: These tests aren't really related to the task at hand (making sure correct createdAt time is returned) but where missing.

They are testing functionality that already existed from the usecase/factory functions.

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -196,7 +196,7 @@ public static CareCaseData ToCareCaseData(this CaseSubmission caseSubmission, Li
                 CaseFormTimestamp = caseSubmission.SubmittedAt?.ToString("yyyy-MM-dd") ?? DateTime.Now.ToString("yyyy-MM-dd"),
                 FormName = caseSubmission.FormId,
                 DateOfBirth = resident.DateOfBirth?.ToString("dd/MM/yyyy"),
-                DateOfEvent = caseSubmission.CreatedAt.ToString("yyyy-MM-dd"),
+                DateOfEvent = caseSubmission.DateOfEvent?.ToString("yyyy-MM-dd") ?? caseSubmission.CreatedAt.ToString("yyyy-MM-dd"),
```
jhackett1, line 199: how will this play with the need to order by time as well as date?

jhackett1 (approved):
i don't see any references to ordering or sorting on the cases endpoint - is that coming later or is that a front-end responsibility? otherwise looks fine to me

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
JohnFarrellDev: In `CaseRecordsUseCase.cs` there is a function `SortData` that handles the sorting in the backend

JohnFarrellDev:
> i don't see any references to ordering or sorting on the cases endpoint - is that coming later or is that a front-end responsibility? otherwise looks fine to me

In `CaseRecordsUseCase.cs` there is a function `SortData` that handles the sorting in the backend

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
JohnFarrellDev: Let me check, if ISO is possible anywhere I always prefer to use it

JohnFarrellDev: Using ISO now which will play better with date and time sorting! (plus should always be parseable)

PR #394, opened 2021-08-02
Disable marked for deletion support for now

LBHTKarki:

Disable the marked for deletion column until the import spreadsheet is in sync. The column must be in the spreadsheet before this is enabled, otherwise the import will fail

Comment out the column and tests 

#### _Checklist_


mapleleafjack (approved):
// 🚀

PR #396, opened 2021-08-04
Development

JohnFarrellDev:
### Changes

- Add title property to a FormSubmission
- Remove tags property from FormSubmission
- Add DateOfEvent to out FormSubmissions
- Importing of people

#### _Checklist_

Deploy!


SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/MongoDBImportTests.cs
```diff
@@ -0,0 +1,112 @@
...
+//            var filterByMarkedForDeletionFlag = Builders<MongoDBTestObject>.Filter.Eq("MarkedForDeletion", "true");
+//            var newMatchingRecords = collection.Find(filterByMarkedForDeletionFlag).ToList();
+//            newMatchingRecords.Count.Should().Be(0);
+
+//            var unchangedRecordsFilter = Builders<MongoDBTestObject>.Filter.Ne("MarkedForDeletion", "true");
+//            var unchangedRecords = collection.Find(unchangedRecordsFilter).ToList();
+//            unchangedRecords.Count.Should().Be(2);
+//        }
+//    }
+//}
```
JohnFarrellDev, line 112: Should we remove a commented out file for now? Large sections of commented code just add confusion for other developers.

SocialCareCaseViewerApi.Tests/V1/Data/ASCEmergencyIdImportTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                                    returning person_id into new_person_mosaic_id;
+
+                                    insert into dbo.sccv_persons_lookup(person_id, nc_id)
+                                            values (new_person_mosaic_id, new_person.person_id);
+
+                                    raise notice '% - %', new_person.person_id, new_person_mosaic_id;
+                            end loop;
+                    end;
+                    $$
+                   ";
```
JohnFarrellDev, line 77: Can we do this without large query strings that are executed as SQL?

I think I understand what the tests here are doing but it definitely makes the tests harder to understand.
If we consider tests as proof our code does what we say it does we want to make our tests as readable as possible.

LBHTKarki: These tests are a bit odd since they test SQL scripts designed to be run manually. They are there just to give more confidence that they actually do what the tests say. Idea is that you should be able to copy and paste that script to SQL server and run it. We are slowly introducing more features in the app to cover these random import/update scripts and we really want to get away from them altogether. There might be better ways to set up the tests, but I couldn't think of any. Happy to review for future scripts to make them easier to manage

SocialCareCaseViewerApi.Tests/V1/Data/MongoDB/MongoDBImportTests.cs
LBHTKarki: I agree, I was hoping to be able to enable that before merging to master, but there have been some delays on the dependencies outside this solution. To keep the master clean it's best to remove that fow now.

SocialCareCaseViewerApi.Tests/V1/Data/ASCEmergencyIdImportTests.cs
JohnFarrellDev: Ok, that makes sense for now :+1:

LBHTKarki (approved):

PR #404, opened 2021-08-06
Add script for updating form ids that are set incorrectly

LBHTKarki:

Due to a bug on the front end all adult case notes have been created with FormId set to child-case-note. This will cause issues with reporting and elsewhere.

Add manual data update script that can be run to update all existing records

#### _Checklist_

Create a plan for updating records manually. This must be done before going live with the CFS case notes to avoid overriding valid values.


JohnFarrellDev (approved):
:+1:

PR #410, opened 2021-08-11
Fix(SCT-983): No title on record history

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-983)

We need to return some extra information about our forms so the frontend can display the title correctly.

Added an optional FormType and Title field in our response.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
@@ -541,33 +541,6 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
 
             response.DateOfEvent.Should().Be("2021-07-18T14:40:30.0000000");
         }
-
-        [Test]
-        public void CaseSubmissionToCareCaseDataReturnsTitleAndFormIdForFormNameWhenTitleProvided()
-        {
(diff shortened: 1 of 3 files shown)


SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
```diff
@@ -541,33 +541,6 @@ public void CaseSubmissionToCareCaseDataReturnsDateOfEventAsCreatedAtIfDateOfEve
 
             response.DateOfEvent.Should().Be("2021-07-18T14:40:30.0000000");
         }
-
-        [Test]
-        public void CaseSubmissionToCareCaseDataReturnsTitleAndFormIdForFormNameWhenTitleProvided()
```
Renny4Real, line 546: Why are removing these tests?

Renny4Real (approved):

SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
JohnFarrellDev: We no longer provide the title in the FormName property, we return an optional title property instead.

PR #414, opened 2021-08-17
Feat(SCT-1089): Create FormValues GET endpoint

acodeninja:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1089

Using an extendable key value style store for case status drop down fields.
Adding an API endpoint that allows the frontend to get fields and options for a case status.

* added key value tables to store field options for case statuses
* removed case status sub type table and moved to new key value tables
* updated infrastructure and tests to reflect the new layout
* added an endpoint at /case-status/form-options/{type} to get available fields and their options.

new structure
image

#### _Checklist_

Run the manual migration listed in the PR prior to rolling to staging and production.


SocialCareCaseViewerApi/V1/Controllers/CaseStatusTypeFieldsController.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <response code="200">Found case status type, returns the fields for that type</response>
+        /// <response code="404">Case status type not found, or no fields exist</response>
+        [ProducesResponseType(typeof(GetCaseStatusFieldsResponse), StatusCodes.Status200OK)]
+        [HttpGet]
+        [Route("/case-status/form-options/{type:alpha}")]
+        public IActionResult GetCaseStatusTypeFields(string type)
+        {
+            try
+            {
+                return Ok(_getCaseStatusFieldsUseCase.Execute(new GetCaseStatusFieldsRequest { Type = type }));
```
LBHTKarki, line 36: This boundary object is perhaps unnecessary if it has only one property. Use case could receive the incoming string as is

SocialCareCaseViewerApi/V1/Controllers/CaseStatusTypeFieldsController.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Get a list of fields and their options for a case status type
+        /// </summary>
+        /// <response code="200">Found case status type, returns the fields for that type</response>
+        /// <response code="404">Case status type not found, or no fields exist</response>
+        [ProducesResponseType(typeof(GetCaseStatusFieldsResponse), StatusCodes.Status200OK)]
+        [HttpGet]
+        [Route("/case-status/form-options/{type:alpha}")]
+        public IActionResult GetCaseStatusTypeFields(string type)
```
LBHTKarki, line 32: Incoming string should probably have validation. If the string is not provided, the API will simply return 404 in this instance. Adding validation would provide correct status code

SocialCareCaseViewerApi/V1/Controllers/CaseStatusTypeFieldsController.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        /// <summary>
+        /// Get a list of fields and their options for a case status type
+        /// </summary>
+        /// <response code="200">Found case status type, returns the fields for that type</response>
+        /// <response code="404">Case status type not found, or no fields exist</response>
+        [ProducesResponseType(typeof(GetCaseStatusFieldsResponse), StatusCodes.Status200OK)]
+        [HttpGet]
+        [Route("/case-status/form-options/{type:alpha}")]
```
LBHTKarki, line 31: leading slash in the route here will result in 404

SocialCareCaseViewerApi/V1/UseCase/GetCaseStatusFieldsUseCase.cs
```diff
@@ -0,0 +1,41 @@
...
+            return new GetCaseStatusFieldsResponse
+            {
+                Description = caseStatusType.Description,
+                Name = caseStatusType.Name,
+                Fields = caseStatusType?.Fields.ToResponse()
+            };
+        }
+    }
+
+    public class CaseStatusNotFoundException : Exception
```
LBHTKarki, line 37: This could perhaps be moved to custom exceptions in exceptions folder so they are all in on place

acodeninja:
Latest CI test fail is down to an issue fixed in https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/418

SocialCareCaseViewerApi/V1/Controllers/CaseStatusTypeFieldsController.cs
acodeninja: Ended up using the request to do validation as the type is now optional in the url.

JohnFarrellDev (approved):
:+1:

PR #416, opened 2021-08-19
fix(SCT-1082): large response size getting submissions

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-1082)

Response size from GET /api/v1/submissions is way too large.

By default do not include EditHistory or FormAnswers when getting a list of case submissions to reduce JSON response size.

#### _Checklist_

Associated frontend changes

### DIFF
--- a/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
+++ b/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
@@ -130,7 +130,7 @@ namespace SocialCareCaseViewerApi.V1.Factories
             };
         }
 
-        public static Domain.CaseSubmission ToDomain(this CaseSubmission caseSubmission, bool includeAllFields = true)
+        public static Domain.CaseSubmission ToDomain(this CaseSubmission caseSubmission, bool includeFormAnswers = true, bool includeEditHistory = true)
         {
             var mapSubmissionStateToString = new Dictionary<SubmissionState, string> {
                 { SubmissionState.InProgress, "In progress" },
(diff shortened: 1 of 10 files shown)


SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -156,14 +156,16 @@ public static Domain.CaseSubmission ToDomain(this CaseSubmission caseSubmission)
...
                     EditTime = e.EditTime,
                     Worker = e.Worker.ToDomain(false)
-                }).ToList(),
+                }).ToList() : null,
                 SubmissionState = mapSubmissionStateToString[caseSubmission.SubmissionState],
-                FormAnswers = caseSubmission.FormAnswers,
-                Title = caseSubmission.Title
+                FormAnswers = includeFormAnswers ? caseSubmission.FormAnswers : null,
+                Title = caseSubmission.Title,
+                LastEdited = caseSubmission.EditHistory.Last().EditTime,
```
LBHTKarki, line 167: Will initial submission have LastEdited value? So the .Last() will always have a value

JohnFarrellDev: I'm tempted to make it nullable but based on FormUseCase#ExecutePost the initially created CaseSubmission does start with a single EditHistory item. .Last() should always return an item.

LBHTKarki:
Would be interesting to investigate doing the field filtering in the db query. That would speed up the data load quite a bit I imagine

JohnFarrellDev:
> Would be interesting to investigate doing the field filtering in the db query. That would speed up the data load quite a bit I imagine

It's a good point, I'll look into if there is a way to make the request to not even ask for the properties not required

martapederiva:
> > Would be interesting to investigate doing the field filtering in the db query. That would speed up the data load quite a bit I imagine
> 
> It's a good point, I'll look into if there is a way to make the request to not even ask for the properties not required

Would love to help on this (DBs ❤️ )

LBHTKarki (approved):

JohnFarrellDev:
> Would be interesting to investigate doing the field filtering in the db query. That would speed up the data load quite a bit I imagine

For now we're not going to do this. Reason is that to get LastEdited property on our domain object we require the EditHistory, to get CompletedSteps on our domain object we require FormAnswers.

It may be possible in the future to make this improvement but for now we don't consider it worth the time required.

PR #417, opened 2021-08-19
SCT-462 Add alerts handler lambda

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-923

So far we have used manually setup python lambda for sending CloudWatch alerts to a Google Room (with SNS). This PR adds new .NET Lambda function to replace the manually created one.

Add new alerts handler project and test project to the solution.

This Lambda is not mean to be attached to API gateway so the setup is slightly different we have for the main Lambdas that serve as API endpoints.

Serverless config has been updated and values added to the parameter store, so we can send alerts to different rooms based on the environment.

The CloudWatch/SNS message is sent to the room as is. This is by design and the idea is that purpose of this lambda is to raise an alert which will trigger the actual (manual) troubleshooting workflow.

#### _Checklist_

Once deployed and tested remove the manually created Lambda.


Renny4Real (approved):
Looks good 🎉

PR #418, opened 2021-08-20
Fix(SCT-1103): Flaky integration tests

acodeninja:
### Jira Ticket

https://hackney.atlassian.net/browse/SCT-1103

When the test suite runs there is an occasional error not related to any of the code changed.

```
A total of 1 test files matched the specified pattern.
  X UpdateWorkerWithNewTeamReturnsTheOnlyTheUpdatedTeam [486ms]
  Error Message:
   Npgsql.PostgresException : 42601: syntax error at or near "Conner"
  Stack Trace:
     at Npgsql.NpgsqlConnector.<>c__DisplayClass160_0.<<DoReadMessage>g__ReadMessageLong|0>d.MoveNext()
```

[Example from CI](https://app.circleci.com/pipelines/github/LBHackney-IT/social-care-case-viewer-api/1674/workflows/5b5e22f4-76da-43ac-8814-002131142bc6/jobs/5029)

This is due to the use of `.ExecuteSqlRaw` with unescaped string interpolation. These helpers throw exceptions when `Faker` generates names like `O'Conner`.

I've removed the raw sql based seeders from the integration test helper and replaced them with calls to `Add()` and `SaveChanges()` in the entity framework.

#### _Checklist_

None

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestSetup.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestSetup.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Data;
+using System.IO;
 using System.Net.Http;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
@@ -16,7 +16,6 @@ namespace SocialCareCaseViewerApi.Tests.V1.IntegrationTests
         private ISccvDbContext _mongoDbContext;
         private MockWebApplicationFactory<TStartup> _factory;
         private NpgsqlConnection _connection;
-        private NpgsqlTransaction _transaction;
         protected HttpClient Client { get; private set; }
         protected DatabaseContext DatabaseContext { get; private set; }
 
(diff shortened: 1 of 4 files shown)


Renny4Real:
@acodeninja Could we check that the test still fails when we remove/comment out the code responsible in the gateway? I encountered some false positives when we used just entity framework (similar to the above) which lead to the decision to try using the raw SQL method.

acodeninja:
@Renny4Real I've tried commenting out the whole body of the `DatabaseGateway.UpdateWorker()` function and all tests I would expect to fail (including the integration tests) are. Is this enough?

Renny4Real:
@acodeninja Oh nice! 
We added this test because of a bug where a worker's teams were not being updated properly and we were trying to recreate it locally. Do any of the integration tests fail if you comment out [lines 531 - 533](https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/11534fa6cdad4eba449c5218596fc55660635b20/SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs#L531) within the gateway, as this should be the fix for the bug?

acodeninja:
I've done some work to actually persist data to the database during integration tests, this should give us more confidence with them and avoid issues around change tracking inside a transaction for tests.

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/IntegrationTestSetup.cs
```diff
@@ -70,5 +66,14 @@ public void AfterAllTests()
         {
             _connection.Dispose();
         }
+
+        private void WipeDatabase()
+        {
+            DatabaseContext.Database.ExecuteSqlRaw("DROP SCHEMA dbo CASCADE");
+            var filePath = TestContext.CurrentContext.WorkDirectory.Replace(Path.Combine("SocialCareCaseViewerApi.Tests", "bin", "Debug", "netcoreapp3.1"), "");
```
acodeninja, line 73: @LBHTKarki I couldn't think of a better way to load in the db schema file. Do you have any ideas?

JohnFarrellDev (approved):
:+1:

PR #419, opened 2021-08-23
fix(SCT-1082): reduce json size/tidy up controllers

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-1082) (Though it potentially helps speed up the response from every single API endpoint)

We want to reduce response size to increase speed of our API call to GET /api/vi/submissions
Current response includes a property even when the value is null, we can just not include the property or null value in the response and reduce the JSON response size.

Screenshot from 2021-08-23 08-43-30

Updated the JSON serializer to remove the property if the value is null from our JSON response.

#### _Checklist_


SocialCareCaseViewerApi/Startup.cs
```diff
@@ -43,7 +34,11 @@ public static void ConfigureServices(IServiceCollection services)
             services
                 .AddMvc()
                 .SetCompatibilityVersion(CompatibilityVersion.Version_3_0)
-                .AddFluentValidation();
+                .AddFluentValidation()
+                .AddJsonOptions(options =>
+                {
+                    options.JsonSerializerOptions.IgnoreNullValues = true;
+                });
```
JohnFarrellDev, line 41: This is the change to make null properties not be returned in our JSON response

Renny4Real: Would the front end need to be aware of which properties are not going to be returned. Would it be expecting a specific object and somehow know what to do if a specific property does not exist?

SocialCareCaseViewerApi/Startup.cs
```diff
@@ -116,22 +111,9 @@ public static void ConfigureServices(IServiceCollection services)
...
-            services.AddTransient<IValidator<PatchWarningNoteRequest>, PatchWarningNoteRequestValidator>();
-            services.AddTransient<IValidator<CreateTeamRequest>, CreateTeamRequestValidator>();
-            services.AddTransient<IValidator<GetTeamsRequest>, GetTeamsRequestValidator>();
-            services.AddTransient<IValidator<CreateCaseSubmissionRequest>, CreateCaseSubmissionRequestValidator>();
-            services.AddTransient<IValidator<UpdateCaseSubmissionRequest>, UpdateCaseSubmissionRequestValidator>();
-            services
-                .AddTransient<IValidator<UpdateFormSubmissionAnswersRequest>, UpdateFormSubmissionAnswersValidator>();
-            services.AddTransient<IValidator<CreatePersonalRelationshipRequest>, CreatePersonalRelationshipRequestValidator>();
-            services.AddTransient<IValidator<QueryCaseSubmissionsRequest>, QueryCaseSubmissionsValidator>();
-
```
Renny4Real, line 134: Why are we removing these? Do we no longer need the validators?

SocialCareCaseViewerApi/V1/Controllers/BaseController.cs
```diff
@@ -1,29 +0,0 @@
...
-                var settings = new JsonSerializerSettings();
-                settings.Formatting = Formatting.Indented;
-                settings.ContractResolver = new CamelCasePropertyNamesContractResolver();
-
-                settings.DateTimeZoneHandling = DateTimeZoneHandling.Utc;
-                settings.DateFormatHandling = DateFormatHandling.IsoDateFormat;
-
-                return settings;
-            };
-        }
```
Renny4Real, line 27: Would changing the super class of our constructors from `BaseController` to `Controller` mean we might lose these JSON settings? Would we want these settings?

Renny4Real (commented):
Looks good. Just a few comments

SocialCareCaseViewerApi/V1/Controllers/BaseController.cs
JohnFarrellDev: As far as I could tell these settings weren't actually implemented.

Initially I tried adding the removal of the null properties here and it had 0 impact on the response.

SocialCareCaseViewerApi/Startup.cs
JohnFarrellDev: I don't believe these are doing anything.

In our constructors we actually the validator and call it against the request there doesn't seem to be a requirement to inject the validator into the Startup class.

PR #425, opened 2021-08-24
Spike/case status domain based gateway

acodeninja:
## Could we slice gateways differently?

Trying to start a conversation around how we slice gateways.

Currently gateways are sliced along storage mechanism lines, this means we effectively have a postgres gateway, a mongodb gateway and a social care platform api gateway. The way we have implemented this leads to the use cases having more responsibility than they maybe should.

Currently a use case must know:

* What data to pull from or changes to make to the system state
* What response the caller is expecting
* How to convert from an infrastructure (ORM) object to a domain object
* Which storage mechanisms are needed to get data (eg: postgres for some data and mongodb for other data)

My suggestion moves some of this responsibility into the gateway, making use cases only aware of which domain objects they need to interact with, and which gateways can make those changes.

This means the gateway now abstracts entity framework away from the use case and the gateway accepts and returns domain objects only.

Created a `CaseStatusGateway` with the `GetCaseStatusesByPersonId` method.
Replaced calls in `CaseStatusesUseCase` to the `DatabaseGateway` with calls to `CaseStatusGateway`.

**Note:** I have not removed the old `GetCaseStatusesByPersonId` method from the `DatabaseGateway` as this wasn't really needed to start the discussion.

#### _Checklist_

None.


LBHTKarki:
@acodeninja One thing we've been wondering about the current gateway design is that even though they are built around data sources and should only have simple methods for CRUD operations against entities, they often end up having some logic or related entities handling in them. With EF it's easy to handle complex relationships when modifying data, so it makes sense to utilise that even though it makes the design look a bit inconsistent in places.

Also the response object varies between gateway methods in a way that sometimes they return boundary objects and sometimes simple domain objects.

Moving the entity handling and the current factory responsibilities to the gateway could simplify the flow and use cases in many places.

I think this setup looks quite neat, but making a move to something that's quite different from the usual design we follow requires wider discussions. Standard design doesn't of course mean it's correct or can't be improved, so it's good to take a closer look at it to see if we could improve it. Consistency for support sake is a factor here as well.

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatusControllerTests.cs
```diff
@@ -41,7 +39,7 @@ public void ListCaseStatuses404WithCorrectErrorMessageWhenPersonIsNotFound()
         {
             _mockCaseStatusesUseCase.Setup(x => x.ExecuteGet(It.IsAny<long>())).Throws(new GetCaseStatusesException("Person not found"));
 
-            var response = _statusTypeController.ListCaseStatuses(123456789) as NotFoundObjectResult;
+            var response = _caseStatusController.ListCaseStatuses(123456789) as NotFoundObjectResult;
```
JohnFarrellDev, line 42: This needs to be as ObjectResult. If our API isn't returning a result with a 404 status code this test will still pass as that StatusCode is cast to 404. (I also question if 404 is the correct status code)

SocialCareCaseViewerApi.Tests/V1/Helpers/Domain/CaseStatusDomainHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace SocialCareCaseViewerApi.Tests.V1.Helpers.Domain
+{
+    public static class CaseStatusDomainHelper
+    {
+        public static CaseStatusType NewCaseStatusType()
+        {
+            return new Faker<CaseStatusType>()
+                .RuleFor(cst => cst.Name, f => f.Lorem.Word())
+                .RuleFor(cst => cst.Description, f => f.Lorem.Sentence())
```
JohnFarrellDev, line 14: We've had issues before with faker causing flaky tests when we don't explicitly state the strength length.

For instance using f.Person.Name instead of f.Random.String2(30). If the string chosen is too long it may not be saved to the database and a test will fail only some of the time.

SocialCareCaseViewerApi/V1/Gateways/ICaseStatusGateway.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using SocialCareCaseViewerApi.V1.Domain;
+
+namespace SocialCareCaseViewerApi.V1.Gateways
+{
+    public interface ICaseStatusGateway
+    {
+        public List<CaseStatus> GetCaseStatusesByPersonId(long personId);
```
JohnFarrellDev, line 8: We should be returning an IEnumerable<CaseStatus>, no need to enforece the structure of the data returned by the gateway when possible

SocialCareCaseViewerApi/V1/Gateways/CaseStatusGateway.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Include(cs => cs.SelectedOptions)
+                .ThenInclude(csso => csso.FieldOption)
+                .ThenInclude(fo => fo.TypeField);
+
+            return caseStatuses.Select(cs => new CaseStatus
+            {
+                Id = cs.Id,
+                Type = cs.Type.Name,
+                StartDate = cs.StartDate.ToString("s"),
+                EndDate = cs.EndDate?.ToString("s"),
```
JohnFarrellDev, line 45: Do we even want to convert these to string format yet? Would a C# datetime object not make more sense as our returned datetime object and be easier for our usecase to do any "business logic" against

SocialCareCaseViewerApi/V1/Gateways/CaseStatusGateway.cs
```diff
@@ -0,0 +1,60 @@
...
+                Fields = cs.SelectedOptions.Select(o => new CaseStatusField
+                {
+                    Name = o.FieldOption.TypeField.Name,
+                    Description = o.FieldOption.TypeField.Description,
+                    SelectedOption = new CaseStatusFieldSelectedOption
+                    {
+                        Name = o.FieldOption.Name,
+                        Description = o.FieldOption.Description
+                    }
+                }).ToList()
```
JohnFarrellDev, line 56: Is this why we don't have an EntityFactory method? I'm ok with this if we want to reduce boilerplate code but consistency within the code base will be good whatever is decided.

SocialCareCaseViewerApi/V1/Gateways/CaseStatusGateway.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public CaseStatusGateway(
+            DatabaseContext databaseContext,
+            IProcessDataGateway processDataGateway,
+            ISystemTime systemTime
+        )
+        {
+            _databaseContext = databaseContext;
+            _processDataGateway = processDataGateway;
+            _systemTime = systemTime;
```
JohnFarrellDev, line 27: Can be dropped from constructor/not passed in anywhere CaseStatusGateway is used.

SocialCareCaseViewerApi/V1/UseCase/CaseStatusUseCase.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public ListCaseStatusesResponse ExecuteGet(long personId)
+        {
+            var person = _databaseGateway.GetPersonByMosaicId(personId) ??
+                         throw new GetCaseStatusesException("Person not found");
+
+            var caseStatus = _caseStatusGateway.GetCaseStatusesByPersonId(person.Id);
+
+            return new ListCaseStatusesResponse { PersonId = personId, CaseStatuses = caseStatus };
```
JohnFarrellDev, line 26: Why not return a List<CaseStatusResponse>? 

The PersonId is redundant to return considering it is part of the request

SocialCareCaseViewerApi/V1/UseCase/CaseStatusUseCase.cs
```diff
@@ -0,0 +1,29 @@
...
+        public CaseStatusUseCase(IDatabaseGateway databaseGateway, ICaseStatusGateway caseStatusGateway)
+        {
+            _databaseGateway = databaseGateway;
+            _caseStatusGateway = caseStatusGateway;
+        }
+
+        public ListCaseStatusesResponse ExecuteGet(long personId)
+        {
+            var person = _databaseGateway.GetPersonByMosaicId(personId) ??
+                         throw new GetCaseStatusesException("Person not found");
```
JohnFarrellDev, line 22: We already have a PersonNotFoundException, would that make more sense to reuse?

SocialCareCaseViewerApi/V1/Domain/CaseStatus.cs
```diff
@@ -10,6 +10,6 @@ public class CaseStatus
         public string? StartDate { get; set; }
         public string? EndDate { get; set; }
         public string? Notes { get; set; }
-        public List<CaseStatusField> Fields { get; set; } = new List<CaseStatusField>();
+        public List<CaseStatusField>? Fields { get; set; }
```
JohnFarrellDev, line 13: Can this ever be null?

JohnFarrellDev (changes requested):
Mostly watch our for how faker may make some tests flaky when you don't use String2 for any string

acodeninja:
Closing as this was a PoC to start a discussion around how we slice gateways.

PR #439, opened 2021-09-02
feat(SCT-1023): Case Status POST endpoint

mapleleafjack:
## Link to JIRA ticket

[link to ticket](https://hackney.atlassian.net/browse/SCT-1023)

## Describe this PR

- Added a new endpoint `/resident/case-status`
- Added tests
- Refactored some of the test classes so the source code is tidier
- Merged Case Status endpoint controller and Case Status Form Types controller in a single Case Status controller
- TestHelper.createPerson now expects a long instead of int

Create a Case Status for a person POST endpoint

#### _Checklist_

#### Follow up actions after merging PR

Case Status endpoint names should be aligned.
They're now currently spelled differently, the correct spell should be "case-statuses"


SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
```diff
@@ -0,0 +1,82 @@
...
+            if (!validationResults.IsValid)
+            {
+                return BadRequest(validationResults.ToString());
+            }
+
+            try
+            {
+                _caseStatusesUseCase.ExecutePost(request);
+
+                return CreatedAtAction("CreatePersonCaseStatus", "Successfully created case status.");
```
JohnFarrellDev, line 69: Instead of "CreatePersonCaseStatus" I would advise utilising nameof(CreateCaseStatus)

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507

Also do we want to return the created Case Status object instead of "Successfully created case status.", if we don't then that string is fine but if the frontend requires the data it can be nice to return it from the POST.

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/CaseStatusCreateTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using SocialCareCaseViewerApi.V1.UseCase.Interfaces;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Controllers.Relationship
+{
+    [TestFixture]
+    public class CaseStatusCreateTests
+    {
+        private CaseStatusController _caseStatusController;
+        private Mock<ICaseStatusesUseCase> _mockCaseStatusesUseCase;
+        // private readonly Fixture _fixture = new Fixture();
```
JohnFarrellDev, line 19: just delete the commented out line

SocialCareCaseViewerApi/V1/UseCase/CaseStatusesUseCase.cs
```diff
@@ -30,5 +31,29 @@ public ListCaseStatusesResponse ExecuteGet(long personId)
 
             return response;
         }
+
+        public void ExecutePost(CreateCaseStatusRequest request)
+        {
+            var persons = _databaseGateway.GetPersonByMosaicId(request.PersonId);
```
JohnFarrellDev, line 37: person, singular

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+            RuleForEach(pr => pr.Fields)
+                .ChildRules(field =>
+                {
+                    field.RuleFor(t => t.Name).NotNull().WithMessage("Field must have a name");
+                    field.RuleFor(t => t.Selected).NotNull().WithMessage("Field selected value must not be empty");
+                });
+
+            RuleFor(x => x.StartDate)
+                .LessThan(DateTime.Now).WithMessage("'start_date' must be in the past");
+
```
JohnFarrellDev, line 61: Does EndDate need to be in the future?

What about a rule ensuring EndDate must be grater than StartDate?

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/CaseStatusCreateTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            response?.StatusCode.Should().Be(400);
+            response?.Value.Should().Be(exceptionMessage);
+        }
+
+        [Test]
+        public void WhenRequestIsValidReturnsSuccessfulResponse()
+        {
+            var request = CaseStatusHelper.CreateCaseStatusRequest();
+            var response = _caseStatusController.CreateCaseStatus(request);
```
JohnFarrellDev, line 89: Should we not be returning the response from mocking the useCase?

Then validate that when we call _caseStatusController.CreateCaseStatus(request) we return the 201 staus code and the value is the appropriate string?

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/CaseStatusCreateTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Controllers;
+using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.UseCase.Interfaces;
+
+namespace SocialCareCaseViewerApi.Tests.V1.Controllers.Relationship
```
JohnFarrellDev, line 12: namespace SocialCareCaseViewerApi.Tests.V1.Controllers.CaseStatus

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/CaseStatusListControllerTests.cs
```diff
@@ -11,9 +11,9 @@
 namespace SocialCareCaseViewerApi.Tests.V1.Controllers
```
JohnFarrellDev, line 11: namespace SocialCareCaseViewerApi.Tests.V1.Controllers.CaseStatus

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,45 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
             return response.FirstOrDefault();
         }
 
+        public Infrastructure.CaseStatus CreateCaseStatus(CreateCaseStatusRequest request)
+        {
+            var caseStatus = new Infrastructure.CaseStatus()
+            {
+                PersonId = request.PersonId,
+                TypeId = (long) request.TypeId,
```
JohnFarrellDev, line 1053: the typeId should be a separate value coming from parameter instead of being mutated onto our request object

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+        [JsonPropertyName("personId")]
+        public long PersonId { get; set; }
+
+        [JsonPropertyName("type")]
+        public string Type { get; set; } = null!;
+        [JsonIgnore]
+        public long? TypeId { get; set; }
+
+        [JsonPropertyName("fields")]
+        public List<CaseStatusRequestField>? Fields { get; set; } = new List<CaseStatusRequestField>() { }!;
```
JohnFarrellDev, line 22: `public List<CaseStatusRequestField> Fields { get; set; } = new List<CaseStatusRequestField>();`

this might be more sensible, by default this says if nothing is passed in the fields property it is set to an empty array. this also means that it isn't nullable, otherwise we need some null checks in other places in the code or risk some null reference exceptions which result in a 500 response

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,45 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+                StartDate = request.StartDate,
+                EndDate = request.EndDate,
+                Notes = request.Notes,
+                CreatedBy = request.CreatedBy
+            };
+
+            _databaseContext.CaseStatuses.Add(caseStatus);
+            _databaseContext.SaveChanges();
+
+            foreach (var optionValue in request.Fields)
```
JohnFarrellDev, line 1063: Unless we change the request object we need a null check here before accessing request.Fields

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,45 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+                .Where(fov => fov.TypeFieldId == field.Id)
+                .FirstOrDefault();
+
+                CaseStatusFieldOption fieldOptions = new Infrastructure.CaseStatusFieldOption()
+                {
+                    StatusId = caseStatus.Id,
+                    FieldOptionId = fieldTypeOption.Id
+                };
+                _databaseContext.CaseStatusFieldOptions.Add(fieldOptions);
+            }
```
JohnFarrellDev, line 1080:
```
foreach (var optionValue in request.Fields)
            {
                var field = _databaseContext.CaseStatusTypeFields
                    .FirstOrDefault(f => f.Name == optionValue.Name);

                var fieldTypeOption = _databaseContext.CaseStatusTypeFieldOptions
                    .Where(fov => fov.Name == optionValue.Selected)
                    .FirstOrDefault(fov => fov.TypeFieldId == field.Id);

                var fieldOptions = new CaseStatusFieldOption
                {
                    StatusId = caseStatus.Id,
                    FieldOptionId = fieldTypeOption.Id
                };
                _databaseContext.CaseStatusFieldOptions.Add(fieldOptions);
            }
```

just a bit tidier way of writing it IMO

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,45 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+
+                var fieldTypeOption = _databaseContext.CaseStatusTypeFieldOptions
+                .Where(fov => fov.Name == optionValue.Selected)
+                .Where(fov => fov.TypeFieldId == field.Id)
+                .FirstOrDefault();
+
+                CaseStatusFieldOption fieldOptions = new Infrastructure.CaseStatusFieldOption()
+                {
+                    StatusId = caseStatus.Id,
+                    FieldOptionId = fieldTypeOption.Id
```
JohnFarrellDev, line 1077: fieldTypeOptions could be null, .FirstOrDefault returns null if nothing is found, how do we want to handle this?

Trying to access .Id of null will end up with a 500 response currently.

SocialCareCaseViewerApi/V1/UseCase/CaseStatusesUseCase.cs
```diff
@@ -30,5 +31,29 @@ public ListCaseStatusesResponse ExecuteGet(long personId)
...
+            if (persons == null) throw new PersonNotFoundException($"'personId' with '{request.PersonId}' was not found.");
+
+            var type = _databaseGateway.GetCaseStatusTypeWithFields(request.Type);
+            var typeDoesNotExist = type == null;
+            if (typeDoesNotExist) throw new CaseStatusTypeNotFoundException($"'type' with '{request.Type}' was not found.");
+
+            request.TypeId = type.Id;
```
JohnFarrellDev, line 44: Instead of mutating the request I'd prefer if we pass this value into .CreateCaseStatus

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            act.Should().Throw<CaseStatusTypeNotFoundException>();
+        }
+
+        [Test]
+        public void WhenCreatedByEmailDoesNotExistThrowsWorkerNotFoundException()
+        {
+            _mockDatabaseGateway.Setup(x => x.GetPersonByMosaicId(It.IsAny<long>())).Returns(TestHelpers.CreatePerson((int) _request.PersonId));
+            _mockDatabaseGateway.Setup(x => x.GetWorkerByEmail(It.IsAny<string>()))
+                .Returns((Worker) null);
```
JohnFarrellDev, line 81: If you want a mock to return null you can just not have a .Returns() call after the setup, No need to cast null as a Worker

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.V1.Infrastructure;
+using SocialCareCaseViewerApi.V1.UseCase;
+using System.Collections.Generic;
+using System;
+
+namespace SocialCareCaseViewerApi.Tests.V1.UseCase.Relationships
```
JohnFarrellDev, line 13: namespace SocialCareCaseViewerApi.Tests.V1.UseCase.CaseStatus

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using SocialCareCaseViewerApi.Tests.V1.Helpers;
+using SocialCareCaseViewerApi.V1.Boundary.Requests;
+using SocialCareCaseViewerApi.V1.Exceptions;
+using SocialCareCaseViewerApi.V1.Gateways;
+using SocialCareCaseViewerApi.V1.Infrastructure;
+using SocialCareCaseViewerApi.V1.UseCase;
+using System.Collections.Generic;
```
JohnFarrellDev, line 10: not being used

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            _mockDatabaseGateway = new Mock<IDatabaseGateway>();
+            _caseStatusesUseCase = new CaseStatusesUseCase(_mockDatabaseGateway.Object);
+
+            _mockDatabaseGateway.Setup(x => x.GetCaseStatusTypeWithFields(_typeInRequest.Name))
+                .Returns(_typeInRequest);
+
+            _request = CaseStatusHelper.CreateCaseStatusRequest(type: _typeInRequest.Name);
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonByMosaicId(It.IsAny<int>()))
+                .Returns(TestHelpers.CreatePerson((int) _request.PersonId));
```
JohnFarrellDev, line 35: Why not update TestHelper.CreatePerson to expect a long for the id? The Person class uses a long for the Id field

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            _mockDatabaseGateway.Setup(x => x.CreateCaseStatus(It.IsAny<CreateCaseStatusRequest>()))
+                .Returns(_caseStatus);
+
+            _mockDatabaseGateway.Setup(x => x.GetWorkerByEmail(It.IsAny<string>()))
+                .Returns(TestHelpers.CreateWorker(email: _request.CreatedBy));
+        }
+
+        [Test]
+        public void CallsDatabaseGatewayToCheckPersonExists()
```
JohnFarrellDev, line 47: Should we test that all the functions are called correctly, considering nothing is returned this is the way to test our function works correctly?

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreateCaseStatusTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+
+        [Test]
+        public void AuditsTheCaseStatus()
+        {
+            var person = TestHelpers.CreatePerson();
+            DatabaseContext.Persons.Add(person);
+            DatabaseContext.SaveChanges();
+
+            var (caseStatusType, caseStatusTypeField, caseOptions) = CaseStatusHelper.SaveCaseStatusFieldsToDatabase(DatabaseContext);
```
JohnFarrellDev, line 79: For unused variables from a tuple you typically use an "_"

`var (caseStatusType, _, _) = CaseStatusHelper.SaveCaseStatusFieldsToDatabase(DatabaseContext);`

SocialCareCaseViewerApi.Tests/V1/Gateways/Database/CreateCaseStatusTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            caseStatus?.PersonId.Should().Be(request.PersonId);
+            caseStatus?.TypeId.Should().Be(caseStatusType.Id);
+            caseStatus?.SelectedOptions[0].FieldOptionId.Should().Be(caseOptions.Id);
+        }
+
+        [Test]
+        public void SetsStartDateToNow()
+        {
+            var fakeTime = new DateTime(2000, 1, 1, 15, 30, 0);
+            _mockSystemTime.Setup(time => time.Now).Returns(fakeTime);
```
JohnFarrellDev, line 54: No need to mock system time for this test, we can delete this line

database/schema.sql
```diff
@@ -597,45 +597,30 @@ CREATE TABLE IF NOT EXISTS dbo.sccv_request_audit(
```
JohnFarrellDev, line 597: I'm not certain but when we update the schema I think we  don't just delete columns from existing tables.

Everything has to be a migration so we would add new lines to alter existing by deleting columns?

The current system doing it manually is awful anyway but I'm pretty sure we don't run the whole script just the changes we make. Maybe verify this with @LBHTKarki

mapleleafjack: Hi John, I was under the impression that the db schema updates had their own place in the `database/manual-updates` folder and that the `schema.sql` file is used only for the test database .

I also wonder at this point what's the purpose of `seeds.sql`, what's is lifecycle?

@LBHTKarki do you have any extra info about this?

LBHTKarki: @mapleleafjack the current process for updating the schema is pretty awful like @JohnFarrellDev said. However it is important to keep writing the schema file using the migration approach i.e. altering tables, columns etc. rather than replacing the table definitions in place.

Writing the changes this way ensures we can always apply the changes to any environment safely without having to touch data (in most cases, there are exceptions of course). Say you add a new column to a table. If the table definition is changed in place and you were to write a manual database update for that in the `database/manual-updates` folder, it would be quite complex since you would have to use a temporary table and migrate the data as well. However, if you use the alter table approach in the scheme file, you can simply run the appropriate alter table commands in each environment to get the schema up to date.

Then it's easy to see when merging PRs what db scripts have to be run. Those can then be copied over to the `database/manual-updates` folder so we have a record of what changes have been applied and when. A lot easier to view them that way than working it out from the repo history.

The schema file is also used in testing to set up the entire database which is a good way to keep the db schema as a whole in check. If we can create the database at any time by running the entire file and it's written in migration style, we know the changes have been applied correctly.

As for the seeds file, that was created so we could easily populate test databases with some basic data for testing the app and setting up local dev environments. It has no other purpose at the moment.

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusRequest.cs
mapleleafjack: Hi John,

There's no need for End Date as of now, it's going to be added/evaluated in the next iterations.
It now has been removed from the POST request

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
mapleleafjack: I sadly can't return the CaseStatus data as I receive a JSON exception about the nested entities (references in the database are nested/recursive and the JSON serialization gives an error)

For now I'll make it return a string

mapleleafjack: return CreatedAtAction("nameof(CreateCaseStatus)", "Successfully created case status.");

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public class CreateCaseStatusRequest
+    {
+        [JsonPropertyName("personId")]
+        public long PersonId { get; set; }
+
+        [JsonPropertyName("type")]
+        public string Type { get; set; } = null!;
+        [JsonIgnore]
+        public long? TypeId { get; set; }
```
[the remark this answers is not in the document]
mapleleafjack: TypeId property removed

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/CaseStatusCreateTests.cs
mapleleafjack: Functions now mocks the ExecutePost function to return a mocked case status + evaluates 201 and returned string

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
mapleleafjack: This is not possible, the reason being the database structure.

The request will have only the type as string as the frontend don't know the ID of that particular type.
The backend will need to load the correct row from the database based on the name and get the ID from there

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
mapleleafjack: Added null check

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
mapleleafjack: Added null check

SocialCareCaseViewerApi/V1/UseCase/CaseStatusesUseCase.cs
mapleleafjack: This was actually not necessary, removed

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            _request = CaseStatusHelper.CreateCaseStatusRequest(type: _typeInRequest.Name);
+
+            _mockDatabaseGateway.Setup(x => x.GetPersonByMosaicId(It.IsAny<int>()))
+                .Returns(TestHelpers.CreatePerson((int) _request.PersonId));
+
+            CaseStatus _caseStatus = CaseStatusHelper.CreateCaseStatus();
+
+            _mockDatabaseGateway.Setup(x => x.CreateCaseStatus(It.IsAny<CreateCaseStatusRequest>()))
+                .Returns(_caseStatus);
```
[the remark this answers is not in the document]
mapleleafjack: Oh good to know!

Changed to use var instead of (Domain or Infrastructure) CaseStatus

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
mapleleafjack: Changed function parameter from int to long

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
mapleleafjack: Added all the functions in this test

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// Create a case status
+        /// </summary>
+        /// <param name="request"></param>
+        /// <response code="201">Successfully created case status</response>
+        /// <response code="400">Invalid CreatePersonCaseStatusRequest received</response>
+        [ProducesResponseType(StatusCodes.Status201Created)]
+        [HttpPost]
+        [Route("resident/case-statuses")]
```
LBHTKarki, line 75: residents resource should probably be plural here

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,55 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+            {
+                PersonId = request.PersonId,
+                TypeId = statusType.Id,
+                StartDate = request.StartDate,
+                Notes = request.Notes,
+                CreatedBy = request.CreatedBy
+            };
+
+            _databaseContext.CaseStatuses.Add(caseStatus);
+            _databaseContext.SaveChanges();
```
LBHTKarki, line 1065: For more robust data changes, would be better to wrap all changes in a single transaction i.e. one SaveChanges() operation. This could be achieved by commenting out this line and making the changes higlighted below.

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,55 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+
+                    if (fieldTypeOption != null)
+                    {
+                        var fieldOptions = new CaseStatusFieldOption
+                        {
+                            StatusId = caseStatus.Id,
+                            FieldOptionId = fieldTypeOption.Id
+                        };
+
+                        _databaseContext.CaseStatusFieldOptions.Add(fieldOptions);
```
LBHTKarki, line 1086: instead of adding the field options separately in their own save, you could add the options to the case status object itself like this: `caseStatus.SelectedOptions.Add(fieldOptions);` because the navigation properties are set correctly EF Core is able to create both entities at the same time. This ensures you'll always either get the complete dataset or nothing at all. When saving them separately there's a risk that one save could fail and you would end up with missing data.

Please ensure 'SelectedOptions' is instantiated first if the above approach is used.

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
mapleleafjack: Adjusting it to plural

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
mapleleafjack: Oh that's useful, didn't know that!

I will adjust the insert accordingly to this :)

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
LBHTKarki: Might be worth looking into additional end point that would return the available types with IDs. Then the front end could pass the ID instead of the name. That way the options would be dynamic for the types as well. Perhaps one for furher iteration.

SocialCareCaseViewerApi/V1/Gateways/DatabaseGateway.cs
```diff
@@ -1032,6 +1045,57 @@ public Infrastructure.CaseStatusType GetCaseStatusTypeWithFields(string type)
...
+                    var field = _databaseContext.CaseStatusTypeFields
+                        .FirstOrDefault(f => f.Name == optionValue.Name);
+
+                    var fieldTypeOption = _databaseContext.CaseStatusTypeFieldOptions
+                        .Where(fov => fov.Name == optionValue.Selected)
+                        .FirstOrDefault(fov => fov.TypeFieldId == field.Id);
+
+                    if (fieldTypeOption != null)
+                    {
+                        var fieldOptions = new CaseStatusFieldOption
```
LBHTKarki, line 1079: `fieldOption` would be tidier name here

mapleleafjack: Changed to singular

LBHTKarki (approved):

PR #444, opened 2021-09-13
bug: handle null mosaic id when requesting cases

JohnFarrellDev:
## Link to JIRA ticket

No ticket

When a null mosaic id was provided we tried to parse an empty string to a long, this throws an unhanded exception.

When null mosaic id provided get the first resident instead of trying to parse empty string.

#### _Checklist_


martapederiva (approved):
🆒

PR #449, opened 2021-09-16
fix: Return pruned response for in progress submissions

JohnFarrellDev:
## Link to JIRA ticket

No ticket.

Reducing the response size/reducing the response time.
Can see the response size difference below, roughly a 2/3rds reduction in response size. Ignore the response time as those requests are running locally, hoping for a 50% reduction in speed which I'll test out of curiosity once this is on staging.

Before:

Screenshot from 2021-09-17 00-24-47

After:

Screenshot from 2021-09-17 00-22-27

Created an optional parameter for pruning our unfinished submission response. This can go into prod before any FE changes are released as the parameter if not provided is defaulted to false.

All we do is if the parameter is set to true in out entity factory/response factory do not return anything not required by the frontend, any property not required is null and therefore not return in the JSON response.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/QueryCaseSubmissionsRequest.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/QueryCaseSubmissionsRequest.cs
@@ -1,45 +1,44 @@
 using System;
 using System.Collections.Generic;
-using System.Text.Json.Serialization;
 using FluentValidation;
+using Microsoft.AspNetCore.Mvc;
 
 namespace SocialCareCaseViewerApi.V1.Boundary.Requests
 #nullable enable
 {
     public class QueryCaseSubmissionsRequest
     {
-        [JsonPropertyName("formId")]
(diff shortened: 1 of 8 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/QueryCaseSubmissionsRequest.cs
```diff
@@ -1,46 +1,48 @@
...
-using System.Text.Json.Serialization;
 using FluentValidation;
+using Microsoft.AspNetCore.Mvc;
 
 namespace SocialCareCaseViewerApi.V1.Boundary.Requests
 #nullable enable
 {
     public class QueryCaseSubmissionsRequest
     {
-        [JsonPropertyName("formId")]
```
JohnFarrellDev, line 11: These were never JSON properties, whoops.
Somehow .net doesn't mind anyway but I'd rather the attribute actually be correct.

PR #456, opened 2021-09-22
fix(SCT-1266, SCT-1267): Optimise and index query for Residents.Firstname and Residents.LastName

mapleleafjack:
## Link to JIRA ticket

[Optimise and index query for Residents.Firstname ](https://hackney.atlassian.net/browse/SCT-1266)
[Optimise and index query for Residents.Lastname](https://hackney.atlassian.net/browse/SCT-1267)

We're optimising the way data is loaded form MongoDB.
The current implementation uses the `ElemMatch` function, it is slow as it iterates over all the columns.

Updated our requests to DocumentDb to use `eq` instead of `elemmatch` (documentdb can not use elemmatch for indexed fields)

Index our case submissions using the resident's first name and last name.

#### _Checklist_


martapederiva (commented):
👍

Renny4Real (approved):

JohnFarrellDev (approved):
:+1:

PR #465, opened 2021-09-28
feat(SCT-1224): patch case status to end case status

JohnFarrellDev:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1224

Ability to end a case status by setting an end date.

Created a PATCH endpoint for case-stauses that allows us to set an end_date on the case status.

#### _Checklist_


mapleleafjack (approved):

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
```diff
@@ -185,7 +186,7 @@ public static InfrastructurePerson CreatePerson(long? personId = null, string? f
                 .RuleFor(p => p.Nationality, f => f.Address.Country())
                 .RuleFor(p => p.NhsNumber, f => f.Random.Number(int.MaxValue))
                 .RuleFor(p => p.PersonIdLegacy, f => f.Random.String2(16))
-                .RuleFor(p => p.AgeContext, f => f.Random.String2(1))
+                .RuleFor(p => p.AgeContext, f => ageContext ?? f.Random.String2(1, "ACac"))
```
Renny4Real, line 189: Why do we use lower case `a` & `c`?

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
```diff
@@ -634,5 +633,32 @@ public static GetTeamsRequest CreateGetTeamsRequest(string? contextFlag = null)
...
         }
+
+        private static List<CaseStatusValue> CreateCaseStatusValues()
+        {
+            var caseStatusValues = new List<CaseStatusValue>();
+
+            for (var i = 0; i < new Random().Next(1, 10); i++)
+            {
+                var value = new Faker<CaseStatusValue>()
+                    .RuleFor(c => c.Name, f => f.Random.String2(1000))
```
Renny4Real, line 644: If these are entered into a separate table for answers, could we run the risk of the `name` or `value` being too large leading to flaky tests?

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
```diff
@@ -43,15 +44,15 @@ public void SetUp()
...
                 .Returns(caseStatus);
-
-            _mockDatabaseGateway.Setup(x => x.GetWorkerByEmail(It.IsAny<string>()))
-                .Returns(TestHelpers.CreateWorker(email: _request.CreatedBy));
         }
 
         [Test]
         public void CallsGatewaysToCheckCaseStatusExists()
         {
-            _mockDatabaseGateway.Setup(x => x.GetPersonByMosaicId(It.IsAny<long>())).Returns(TestHelpers.CreatePerson(_request.PersonId));
```
Renny4Real, line 54: Not necessarily for this but wondering if we should rename `GetPersonByMosaicId` to not reference Mosaic at all. What do you think? 🤔

SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateCaseStatusRequest.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        [JsonPropertyName("value")]
+        public string Value { get; set; } = null!;
+    }
+
+    public class UpdateCaseStatusValidator : AbstractValidator<UpdateCaseStatusRequest>
+    {
+        public UpdateCaseStatusValidator()
+        {
+            RuleFor(x => x.EndDate)
```
Renny4Real, line 40: I'm not sure but I think that `end date` could also be in the past but must be after the start date. @mapleleafjack Is this correct?

SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateCaseStatusRequest.cs
```diff
@@ -0,0 +1,55 @@
...
+                .NotNull().WithMessage("'editedBy' must be provided")
+                .EmailAddress().WithMessage("'editedBy' must be a valid email address");
+            RuleFor(x => x.Notes)
+                .MaximumLength(1000).When(x => x.Notes != null).WithMessage("'notes' must be less than or equal to 1,000 characters.");
+            RuleForEach(x => x.Values)
+                .ChildRules(value =>
+                {
+                    value.RuleFor(x => x.Name).NotNull().WithMessage("Field must have a name");
+                    value.RuleFor(x => x.Value).NotNull().WithMessage("Field selected value must not be empty");
+                });
```
Renny4Real, line 52: Would these rules mean you always have to pass in a `CaseStatusValue`? 
Would we need a `.When(x => x.Values != null)` on line 47?

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
```diff
@@ -100,5 +101,44 @@ public IActionResult CreateCaseStatus([FromBody] CreateCaseStatusRequest request
...
+            }
+
+            try
+            {
+                var caseStatus = _caseStatusesUseCase.ExecuteUpdate(caseStatusId, request);
+                return Ok(caseStatus);
+            }
+            catch (Exception e) when (
+                e is PersonNotFoundException ||
+                e is WorkerNotFoundException ||
```
Renny4Real, line 132: Would the Not found exceptions be `404` s? (I don't have strong opinions on this, just interested in learning about status codes in general 🙂 )

Renny4Real (commented):
Looks good 🎉 .

SocialCareCaseViewerApi.Tests/V1/Boundary/Request/UpdateCaseStatusRequestTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+#nullable enable
+namespace SocialCareCaseViewerApi.Tests.V1.Boundary.Request
+{
+    [TestFixture]
+    public class UpdateCaseStatusRequestTests
+    {
+        private static readonly Faker _faker = new Faker();
+
+        private static object[] _updateCaseStatusRequests =
```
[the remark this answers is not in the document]
JohnFarrellDev: Correct, it is a 2 dimensional array where the inner array is object[] (so can hold any element essentially)

SocialCareCaseViewerApi.Tests/V1/Boundary/Request/UpdateCaseStatusRequestTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            new object?[] { TestHelpers.CreateUpdateCaseStatusRequest(), true, ""},
+            new object?[] { TestHelpers.CreateUpdateCaseStatusRequest(DateTime.Now.AddDays(-1)), false, "'endDate' must be in the future"},
+            new object?[] { TestHelpers.CreateUpdateCaseStatusRequest(email: "invalid-email-address"), false, "'editedBy' must be a valid email address"},
+            new object?[] { TestHelpers.CreateUpdateCaseStatusRequest(notes: _faker.Random.String2(1001)), false, "'notes' must be less than or equal to 1,000 characters."},
+        };
+
+        [TestCaseSource(nameof(_updateCaseStatusRequests))]
```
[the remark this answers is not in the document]
JohnFarrellDev: If you run the test you'll see that the test is actually run multiple times.

It is run for every inner array from _updateCaseStatusRequests.

You'll also notice the test takes 3 parameters. these are the 3 elements coming from the inner arrays.

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/UpdateCaseStatusControllerTest.cs
```diff
@@ -0,0 +1,114 @@
...
+            _mockCaseStatusesUseCase
+                .Setup(x => x.ExecuteUpdate(resident.Id, request))
+                .Throws(new PersonNotFoundException(errorMessage));
+
+            var response = _caseStatusController.UpdateCaseStatus(resident.Id, request) as ObjectResult;
+
+            response?.StatusCode.Should().Be(400);
+            response?.Value.Should().Be(errorMessage);
+        }
+
```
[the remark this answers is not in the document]
JohnFarrellDev: I would do a 404 if the aim of the request was to fetch the resident and return and the request did not find a resident.

In this case I prefer a 400 as the aim of the request is to update a case status, if the request is trying to update a case status for a person that does not exist I consider it a bad request.

SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateCaseStatusRequest.cs
JohnFarrellDev: Nah, you could just not pass in a list of Values or even an empty list.
The ChildRules is saying that for every item in the list of Values you must provide a Name and Value.

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
JohnFarrellDev: I consider if the worker/resident isn't found then the request to update the CaseStatus was a Badrequest. 
The resident/worker is not the primary focus of the request so I only return a 404 when the actual CaseStatus isn't found.

SocialCareCaseViewerApi.Tests/V1/UseCase/CaseStatus/CaseStatusExecutePostUseCaseTests.cs
JohnFarrellDev: Agreed, I'll do this in a separate PR to not grow this PR too large.

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
JohnFarrellDev: It is probably quite nice if we are able to ignore the case for A or C from what the frontend sends.
In our actual code we .ToUpper I think.
So if our TestHelpers use either case it hopefully enforces that our code  is handling either case.

SocialCareCaseViewerApi.Tests/V1/Gateways/CaseStatusGatewayTests/GetCaseStatusByCaseStatusIdTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        [Test]
+        public void GetCaseStatusByIdReturnsTheAssociatedCaseStatus()
+        {
+            var (caseStatus, _) = CaseStatusHelper.SavePersonWithCaseStatusToDatabase(DatabaseContext);
+
+            var response = _caseStatusGateway.GetCasesStatusByCaseStatusId(caseStatus.Id);
+
+            CaseStatusHelper.TrimMilliseconds(response?.StartDate).Should().Be(CaseStatusHelper.TrimMilliseconds(caseStatus.ToDomain().StartDate));
```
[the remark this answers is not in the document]
JohnFarrellDev: Possibly. I imagine we don't care to the millisecond the startDates and endDates etc.

This bit of code was to fix the test which for some reason (I couldn't figure out) was being out by a fraction of a ms.
If I change the code to always set the ms to 0 when saving the data then I can remove the TrimMilliseconds from the test

SocialCareCaseViewerApi.Tests/V1/Gateways/CaseStatusGatewayTests/GetCaseStatusByCaseStatusIdTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    [TestFixture]
+    public class GetCaseStatusByCaseStatusIdTests : DatabaseTests
+    {
+        private CaseStatusGateway _caseStatusGateway = null!;
+
+        [SetUp]
+        public void Setup()
+        {
+            _caseStatusGateway = new CaseStatusGateway(DatabaseContext);
+            DatabaseContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
```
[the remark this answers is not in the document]
JohnFarrellDev: https://docs.microsoft.com/en-us/ef/core/querying/tracking

I'm pretty sure NoTracking just means the entity framework will not actually save the changes to the database.

SocialCareCaseViewerApi.Tests/V1/Controllers/CaseStatus/UpdateCaseStatusControllerTest.cs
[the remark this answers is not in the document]
Renny4Real: I see. Thanks!

SocialCareCaseViewerApi.Tests/V1/Gateways/CaseStatusGatewayTests/GetCaseStatusByCaseStatusIdTests.cs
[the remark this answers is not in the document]
Renny4Real: Cool. 🎉  
Yeah, I was thinking if we're removing milliseconds when we get the data, should we be saving the data with milliseconds but I understand the time difference weirdness that could happen while testing. So trimming does makes sense.

SocialCareCaseViewerApi.Tests/V1/Helpers/TestHelpers.cs
Renny4Real: I see. Thanks! 🙏🏾

SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateCaseStatusRequest.cs
Renny4Real: So as long as there is no item, the child rules don't apply. But if there is an item in the array then the validation kicks in.

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
Renny4Real: That makes sense. 🙌🏾 . Thanks!

SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateCaseStatusRequest.cs
mapleleafjack: That's right, `end date` can be either in the past or future, but not before the start date of the Case Status!

The UI already prevents to send a date past the Case Status start but having a check here will probably be useful :)

PR #478, opened 2021-10-14
(SCT-1310) feature - POST end point for adding case status answer for LAC

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1310

We need an endpoint that let us record changes to child's circumstances

Add new POST end point that allows us to add new answers to the case status which are used to indicate changes in circumstances.

Only LAC status can have these additional answers, so requests to add answers to other types will be rejected.

Also tidied up the unrelated CreateCaseStatusRequestTest.cs file.

#### _Checklist_


SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusAnswerRequest.cs
```diff
@@ -0,0 +1,48 @@
...
+            RuleFor(pr => pr.Answers).Must(x => x.Count > 0).WithMessage("Answers must be provided");
+
+            RuleForEach(pr => pr.Answers)
+                .ChildRules(field =>
+                {
+                    field.RuleFor(t => t.Option).NotEmpty().WithMessage("Option must not be empty");
+                    field.RuleFor(t => t.Value).NotEmpty().WithMessage("Value must not be empty");
+                });
+
+            RuleFor(x => x.StartDate)
```
mapleleafjack, line 40: I think Case Status Answers should be both in the future and in the past, as we can have future pending status changes

SocialCareCaseViewerApi.Tests/V1/Boundary/Request/CreateCaseStatusAnswersRequestTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var response = _createCaseStatusAnswerRequestValidator.Validate(requestWithoutAnswers);
+
+            response.IsValid.Should().BeFalse();
+            response.Errors.Should().Contain(e => e.ErrorMessage == "Option must not be empty");
+            response.Errors.Should().Contain(e => e.ErrorMessage == "Value must not be empty");
+        }
+
+        [Test]
+        public void WhenStartDateIsInTheFutureReturnsErrorWithMessage()
```
mapleleafjack, line 54: I think Case Status Answers should be both in the future and in the past, as we can have future pending status changes

SocialCareCaseViewerApi/V1/Controllers/CaseStatusController.cs
```diff
@@ -81,6 +81,42 @@ public IActionResult CreateCaseStatus([FromBody] CreateCaseStatusRequest request
...
+                return CreatedAtAction(nameof(CreateCaseStatusAnswers), "Successfully created case status answers.");
+            }
+            catch (Exception e) when (
+                e is WorkerNotFoundException)
+            {
+                return BadRequest(e.Message);
+            }
+            catch (CaseStatusDoesNotExistException e)
+            {
+                return NotFound(e.Message);
```
JohnFarrellDev, line 116: My only suggestion is that I don't think a POST endpoint should be returning a 404, if the Case Status does not exist then the request was bad and a 400 seems appropriate.

SocialCareCaseViewerApi/V1/Boundary/Requests/CreateCaseStatusAnswerRequest.cs
LBHTKarki: @mapleleafjack  That's true. I've pushed an update to allow both future and past dates

PR #508, opened 2021-11-03
feat(SCT-1424): referral screening decision

JohnFarrellDev:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-1424)

API needs to accept a patch request that can update a mash referral from the screening stage to the final stage and store the submitted screening decisions.

Made patch endpoint.
Validate if screening update the request contains other required screening fields.
Update the referral with those decisions and save changed back to db.

#### _Checklist_

Minor changes to FE for MashReferral interface.

### DIFF
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateMashReferral.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateMashReferral.cs
@@ -1,0 +1,26 @@
+
+using FluentValidation;
+
+#nullable enable
+namespace SocialCareCaseViewerApi.V1.Boundary.Requests
+{
+    public class UpdateMashReferral
+    {
+        public string UpdateType { get; set; }
+        public string? Decision { get; set; }
+        public bool? RequiresUrgentContact { get; set; }
+    }
+
+    public class UpdateMashReferralValidator : AbstractValidator<UpdateMashReferral>
+    {
+        public UpdateMashReferralValidator()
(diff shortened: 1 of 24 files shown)


SocialCareCaseViewerApi/V1/Boundary/Requests/UpdateMashReferral.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class UpdateMashReferralValidator : AbstractValidator<UpdateMashReferral>
+    {
+        public UpdateMashReferralValidator()
+        {
+            RuleFor(x => x.Decision)
+                .NotNull().When(x => x.UpdateType.ToUpper() == "SCREENING-DECISION")
+                .WithMessage("Must provide a decision");
+            RuleFor(x => x.Decision)
+                .MinimumLength(1).When(x => x.UpdateType.ToUpper() == "SCREENING-DECISION")
+                .WithMessage("Must provide a decision");
```
martapederiva, line 24: out of curiosity, what is the difference between a null value and a value that has a <1 length? isn't the latter null by definition? 🤔

JohnFarrellDev: null is to represent nothing being sent, an empty string ("") is not null but would also not be something we want to accept.

Miles-Alford (approved):
LGTM

PR #511, opened 2021-11-04
feat(SCT-1440): MASH Postgres data structure

martapederiva:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1440

We need to save the MASH data to Postgres as well as to MongoDB
This PR is to create the tables for it

Two new tables, one for the referrals and one for the residents involved

#### _Checklist_


database/schema.sql
```diff
@@ -842,3 +842,47 @@ ALTER TABLE dbo.sccv_person_case_status_answers
...
 	ADD COLUMN end_date timestamp;
+
+  --MASH tables
+  CREATE TABLE IF NOT EXISTS dbo.REF_MASH_REFERRALS
+(
+  id bigint GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
+  referrer varchar(256),
+  requested_support varchar(256),	
+  worker_id bigint,
+  created_at timestamp without time zone,
```
JohnFarrellDev, line 853: Are you sure we don't care about timezone? Do we need to know the timezone because of the clocks going forward and back one hour for BST/UTC?

database/schema.sql
```diff
@@ -842,3 +842,47 @@ ALTER TABLE dbo.sccv_person_case_status_answers
...
+  stage varchar(30),
+  referral_category varchar(256),
+  initial_decision varchar(256),
+  initial_decision_referral_category varchar(256),
+  initial_decision_urgent_contact bool,
+  screening_decision varchar(256),
+  screening_decision_urgent_contact bool,
+  final_decision varchar(256),
+  final_decision_referral_category varchar(256),
+  final_decision_urgent_contact bool,
```
JohnFarrellDev, line 864: initial, screening and final all need a created_at, so initial_decision_created_at which in the code will be a DateTime, so a timestamp would make sense I expect as the type

database/schema.sql
martapederiva: as there is no compelling reason for excluding the timezone, will keep as regular timestamp.

database/schema.sql
martapederiva: will add the new fields in upcoming commit

PR #513, opened 2021-11-05
SCT-1432 Add support for closed statuses

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1432

Current get end point only returns active case statuses. We need to be able to get closed statuses as well so we can improve the date validation on the front end.

Add `include_closed_cases` query string parameter which can be set to true if closed cases should be returned as well.

Integration test is there to check that the get end point is setup correctly to handle `personId` from the route and the `include_closed_cases` flag from the query string.

#### _Checklist_


rachel-hutchinson (approved):
Looks good to me

PR #516, opened 2021-11-08
SCT-1450 update start date handling when patching without end date

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1450

Currently, the backend is configured so that if we are editing LAC we update the start date of the `dbo.sccv_person_case_status` table only if there is one group of answers.

However, there could be the following scenario where we would still want to edit the start in the `dbo.sccv_person_case_status` table (and there would be two groups of answers):

1. Add a LAC status (starting 01/11/2021)
3. Update the LAC status so there is now a 'Scheduled status' (scheduled for 01/03/2022)
5. Edit the LAC status (with a new start date of 03/11/2021)

At this point we would still want to update the `dbo.sccv_person_case_status` start date as we are technically still editing the 'first group' of answers 

We would want to end up with the start date of `dbo.sccv_person_case_status` to be 03/11/2021 and the start date of the new answer group (copied/replaced original answers) should also now be 03/11/2021

Update the logic on the gateway so it can handle the scenario above correctly.

#### _Checklist_


mapleleafjack (approved):
Looks good to me!

PR #525, opened 2021-11-15
SCT-1455 Edit case status tests

rachel-hutchinson:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1455

When a LAC case status is updated so that there is a scheduled status, an end date is added to the current case status (this end date is the start date of the scheduled case status).
If the current case status is then edited, the end date was not being copied to the new set of case status answers.

The code was updated so that, if there was an end date for the current case status answers, this was copied to the new set of case status answers.
Additional tests were added to test this.
Following this change, an update was made to the logic used to identify previous case status answers, this was to ensure only previous answers were picked and current case status answers which had an end date were not being modified
Lastly, one test was creating either 1 or 2 answers and this has been changed so that it always creates 2 answers.

#### _Checklist_

It's worth running the frontend tests following the merge of this PR to ensure they all still work as expected


mapleleafjack (approved):
🚀

PR #528, opened 2021-11-18
Changes to dummy data

rs2416:
Some changes to make the dummy data more representative. This includes replacing the names of dummy children from e.g. c1 to realistic names and replacing initial decision inputs so they are more representative.


JohnFarrellDev (approved):
looks good

PR #532, opened 2021-11-22
Feat: fetch worker allocations by worker email

acodeninja:

The allocations endpoint only allows searching by Mosaic ID or Worker ID, this means some actions require multiple HTTP requests to get allocations from a worker's email address. This PR addresses this.

* Add worker_email to List Allocations request object.
* Add logic for fetching allocations by email address to the DB Gateway.

#### _Checklist_

Add worker_email as an option to the swagger documentation.

### DIFF
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
@@ -11,6 +11,9 @@ namespace SocialCareCaseViewerApi.V1.Boundary.Requests
 
         [FromQuery(Name = "worker_id")]
         public long WorkerId { get; set; }
+
+        [FromQuery(Name = "worker_email")]
+        public string WorkerEmail { get; set; }
     }
 
     public class ListAllocationsRequestValidator : ValidationAttribute
--- a/SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
+++ b/SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
@@ -1,11 +1,13 @@
-using System.ComponentModel.DataAnnotations;
+using FluentValidation;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.AspNetCore.Mvc;
 
 namespace SocialCareCaseViewerApi.V1.Boundary.Requests
 {
     public class ListAllocationsRequest
     {
-        [ListAllocationsRequestValidator]
         [FromQuery(Name = "mosaic_id")]
         public long MosaicId { get; set; }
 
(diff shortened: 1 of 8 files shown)


jhackett1 (approved):
this seems to do what i need

mapleleafjack (approved):
LGTM

SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
```diff
@@ -11,6 +11,9 @@ public class ListAllocationsRequest
 
         [FromQuery(Name = "worker_id")]
         public long WorkerId { get; set; }
+
+        [FromQuery(Name = "worker_email")]
```
LBHTKarki, line 15: The way the boundary object validation was initially setup means that providing only `worker_email` will result in 400 error, so minor tweaks are required for the request validator.

acodeninja: I've done my best here, used Fluent validations with two custom validators (couldn't find ones that matched) for "At least one of these properties must be set" and "only one of these properties can be set".

JohnFarrellDev: Might be nice to make the namespace nullable, then instead of assuming defaults (0) as no property passed we'd actually be checking for null/!null values

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
```diff
@@ -42,6 +42,14 @@ public class SocialCareCaseViewerApiController : Controller
         [Route("allocations")]
         public IActionResult GetAllocations([FromQuery] ListAllocationsRequest request)
         {
+            var validator = new ListAllocationsRequestValidator();
+            var validationResults = validator.Validate(request);
+
+            if (!validationResults.IsValid)
+            {
+                return BadRequest(validationResults.ToString());
```
JohnFarrellDev, line 50: code comment above does not match http status response type

JohnFarrellDev:
There does feel an opportunity to refactor was we go here. A new class for AllocationsController and AllocationsGateway would help reduce the code in our controller and gateway mega classes

SocialCareCaseViewerApi/V1/Controllers/SocialCareCaseViewerApiController.cs
acodeninja: Yeup, looking at it the comment was wrong anyways, looks like we return an empty list if there were no allocations found, replacing with just 200 and 400 response codes.

SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
acodeninja: > Might be nice to make the namespace nullable, then instead of assuming defaults (0) as no property passed we'd actually be checking for null/!null values

Do you have an example of this in action?

JohnFarrellDev:
```
#nullable enable
namespace SocialCareCaseViewerApi.V1.Boundary.Requests
{
    public class ListAllocationsRequest
    {
        [FromQuery(Name = "mosaic_id")]
        public long? MosaicId { get; set; }

        [FromQuery(Name = "worker_id")]
        public long? WorkerId { get; set; }

        [FromQuery(Name = "worker_email")]
        public string? WorkerEmail { get; set; }
    }

    public class ListAllocationsRequestValidator : AbstractValidator<ListAllocationsRequest>
    {
        public ListAllocationsRequestValidator()
        {
            RuleFor(x => x)
                .Must(OneIsSet)
                .WithMessage("Please provide either mosaic_id, worker_id or worker_email");

            RuleFor(x => x)
                .Must(OnlyOneIsSet)
                .WithMessage("Please provide only one of mosaic_id, worker_id or worker_email");

            When(x => !String.IsNullOrEmpty(x.WorkerEmail), () =>
            {
                RuleFor(x => x.WorkerEmail)
                    .EmailAddress()
                    .WithMessage("Please provide a valid email address for worker_email");
            });
        }

        private bool OnlyOneIsSet(ListAllocationsRequest request)
        {
            return new List<int>
            {
                request.MosaicId == null ? 0 : 1,
                request.WorkerId == null ? 0 : 1,
                String.IsNullOrEmpty(request.WorkerEmail) ? 0 : 1,
            }.Sum() <= 1;
        }

        private bool OneIsSet(ListAllocationsRequest request)
        {
            return !(request.MosaicId == null && request.WorkerId == null && String.IsNullOrEmpty(request.WorkerEmail));
        }
    }
```

JohnFarrellDev: I've found being explicit about what can and can't be null has resulted in a substantial decrease in server side errors

acodeninja: Ok, cool, I'll get that change in, will save the refactoring for another task because of time pressure.

JohnFarrellDev (approved):
:+1:

SocialCareCaseViewerApi/V1/Boundary/Requests/ListAllocationsRequest.cs
jhackett1: are we happy to merge as is for now?

LBHTKarki (approved):

PR #535, opened 2021-11-25
Add soft delete support for case notes

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1457

As a manager/senior staff member I need to be able to delete system generated records based on specific requests

This adds support for soft deletion of specific submission types.

1. Add new `DELETE` end point that takes the following parameters:
`submissionId` (from route): id of the submission to be deleted
`deletedBy`: email of the worker deleting the record
`deleteReason`: reason why the record has been requested to be deleted
`deleteRequestedBy`: name of the person requesting the deletion
2. End point returns the following status codes:
`204`: Case submission successfully deleted
`400`: Invalid DeleteCaseSubmissionRequest received
`422`: Case submission not found or worker not found
`410`: Case submission already deleted
3. Added new generic mongo gateway method `GetRecordsCountByFilter`
 This is required so we can get the deleted records count on every GET request even though the frontend haven't requested the actual deleted records to be returned. This is so the front end can show that there are deleted records available for the given query filter
4. Updated the `Paginated` object to include `DeletedItemsCount` property for passing the deleted records count to the front end
5.  Updated `GenerateFilter` method in the `FormSubmissionUseCase` to support filtering by `Deleted` property
6.  Updated test helpers to support new `Deleted` and `DeletionDetails` properties
7.  Updated EntityFactory to support `Deleted` and `DeletionDetails` properties
8.  Updated `GetSubmissionByQueryParameters` end point to support `includeDeletedRecords` boolean parameter. This is set to  false by default
9.  Updated `CaseSubmissionResponse` boundary object to include `IsImported` boolean property. This is not used anywhere else, but is there purely for the purpose of making it possible for the front end to identify whether the record was created through the app or was imported. Submission end points do not handle imported records at all, so this can be left in place and changed if there ever is the need to support that.
10. Updated `CaseSubmission` domain object to support `Deleted` and `DeletionDetails` properties
11. Added custom exceptions
12. Updated `CaseSubmission` infrastructure object to support the following properties:
 `Deleted`, nullable boolean with default value false. Any existing submission record won't have this property, so making it nullable with default value ensures all existing records can be handled without issues and that the front end always has that flag in place when applying business logic
 `DeletionDetails `, nullable. If the existing submission record don't have this value, it won't get returned to the front end
 All new records will have these two values, but only `Deleted` property will be included in the response object by default
13. Added new `ExecuteDelete` method to `FormSubmissionsUseCase` 
 only allows the deletion of submissions that have formId set to either `child-case-note` or `adult-case-note`. Supported form types are handled in the static `FormIdName` class to reduce the amount of magic strings in the code

This update is desigend to handle soft deletion of submissions in the `resident-case-submissions` collection and limited to specific formId types mentioned above.

To add the same support for records in `form_data` collection will require further development work and is out of scope of this update.

#### _Checklist_


SocialCareCaseViewerApi/V1/UseCase/FormSubmissionsUseCase.cs
```diff
@@ -67,7 +67,41 @@ public FormSubmissionsUseCase(IDatabaseGateway databaseGateway, IMongoGateway mo
...
+            {
+                throw new DeleteSubmissionException($"Submission with ID {submissionId} not found");
+            }
+
+            if (submissionToBeDeleted.Deleted == true)
+            {
+                throw new SubmissionAlreadyDeletedException($"Submission with ID { submissionId} has already been deleted");
+            }
+
+            if (submissionToBeDeleted.FormId?.ToLower() != FormIdName.ChildCaseNote && submissionToBeDeleted.FormId?.ToLower() != FormIdName.AdultCaseNote)
```
JohnFarrellDev, line 86: Should this be an or?

Won't this always be false?

SocialCareCaseViewerApi/V1/UseCase/FormSubmissionsUseCase.cs
```diff
@@ -124,6 +158,18 @@ public static FilterDefinition<CaseSubmission> GenerateFilter(QueryCaseSubmissio
                 filter &= MongoDB.Bson.Serialization.BsonSerializer.Deserialize<BsonDocument>(bsonQuery);
             }
 
+            if (addDeletedRecordsFilter)
+            {
+                if (request.IncludeDeletedRecords)
+                {
+                    filter &= (Builders<CaseSubmission>.Filter.Eq(s => s.Deleted, true) | Builders<CaseSubmission>.Filter.Ne(s => s.Deleted, true));
```
JohnFarrellDev, line 165: Worth double checking but I think the way &= for filter works is "subtractive", if you want to have all records (deleted or not) you don't need to add anything to the filter, by default they should both be included.

SocialCareCaseViewerApi/V1/Helpers/Paginated.cs
```diff
@@ -6,5 +6,6 @@ public class Paginated<T>
     {
         public List<T> Items { get; set; }
         public long Count { get; set; }
+        public long DeletedItemsCount { get; set; }
```
JohnFarrellDev, line 9: Instead of modifying this generic class could we extend it?

I'm only asking this because I know this class will be used elsewhere but in all cases we won't have DeletedItems

SocialCareCaseViewerApi/V1/Boundary/Response/CaseSubmissionResponse.cs
```diff
@@ -35,6 +35,10 @@ public class CaseSubmissionResponse
 
         public int? CompletedSteps { get; set; }
 
+        public bool Deleted { get; set; } = false;
+        public DeletionDetails? DeletionDetails { get; set; }
+        public bool IsImported { get; set; } = false;
```
JohnFarrellDev, line 40: this is never set?

SocialCareCaseViewerApi/V1/Boundary/Requests/DeleteCaseSubmissionRequest.cs
```diff
@@ -0,0 +1,30 @@
+using FluentValidation;
+using System.Text.Json.Serialization;
+
```
JohnFarrellDev, line 3: I think we would benefit if we #enable nullable here

SocialCareCaseViewerApi.Tests/V1/Controllers/FormSubmissionControllerTests.cs
```diff
@@ -256,5 +256,76 @@ public void GetSubmissionByQueryParametersReturns400WhenQueryCaseSubmissionsExce
...
         }
+
+        [Test]
+        public void DeleteSubmissionReturns204WhenASubmissionIsSuccessfullyDeleted()
+        {
+            var request = TestHelpers.DeleteCaseSubmissionRequest();
+            var deletedSubmission = TestHelpers.CreateCaseSubmission();
+            _submissionsUseCaseMock.Setup(x => x.ExecuteDelete(deletedSubmission.SubmissionId.ToString(), request)).Verifiable();
+
+            var response = _formSubmissionController.DeleteSubmission(deletedSubmission.SubmissionId.ToString(), request) as NoContentResult;
```
JohnFarrellDev, line 267: For all of these tests we need to do as ObjectResult.

The reason is I think the StatusCode gets cast when you specify the type of ObjectResult.

For instance if you change this to NotFoundObjectResult I expect the test would fail and start saying the StatusCode is 404.

As ObjectResult allows us to access the data without any casting occuring.

JohnFarrellDev (approved):
Just some minor suggestions

SocialCareCaseViewerApi/V1/UseCase/FormSubmissionsUseCase.cs
LBHTKarki: It kind of feels a bit backwards, but using `OR` would mean that when formId is `child-case-note` for example, then 
submissionToBeDeleted.FormId?.ToLower() != `child-case-note` is `false`
submissionToBeDeleted.FormId?.ToLower() != `adult-case-note` is `true`
=> `true`, throws exception

anything other than accepted values will result in `true` && `true` => exception

SocialCareCaseViewerApi/V1/UseCase/FormSubmissionsUseCase.cs
LBHTKarki: I had a look and that's true. I've updated this so the logic works that way. Nice!

SocialCareCaseViewerApi/V1/Helpers/Paginated.cs
LBHTKarki: That's fine, I have added a new class that inherits from this one. Not sure about the name, not very imaginative. I tried to keep that quite generic as well, but happy to rename.

```
public class PaginatedExtended<T> : Paginated<T>
{
	public long DeletedItemsCount { get; set; }
} 
```

this inheritance makes the json properties to appear in funny order i.e. the extended class' properties come first. I don't think that matters as far as code handling is concerned.

SocialCareCaseViewerApi/V1/Boundary/Response/CaseSubmissionResponse.cs
LBHTKarki: I'm not a big fan of adding that "dummy" IsImported property here, but it's there to inform the front end what records can be deleted (we shouldn't touch imported records). The way the front end combines records from submissions and cases end points makes this a bit awkward, but since we are not going to support cases deletion for now, this placeholder property should help the front end with business logic. At least it's not polluting the records themselves and can be removed or develop further if need be.

SocialCareCaseViewerApi/V1/Boundary/Requests/DeleteCaseSubmissionRequest.cs
LBHTKarki: yes, that should be the default setup. Added

SocialCareCaseViewerApi.Tests/V1/Controllers/FormSubmissionControllerTests.cs
LBHTKarki: This is interesting because if `ObjectResult` is used with `NoContent()` or `StatusCode(204)` the response is always null. I can't see how the controller setup could be changed to make that `as ObjectResult` work in the test.

I agree ObjectResult should be used, but looks like 204 is a special case that works a bit differently from other codes we normally use. 

I've changed one of the tests to use ObjectResult that should've had it.

As for your comment about changing the above to `NotFoundObjectResult`, that results in null response as well. It seems the conversion fails with anything other than `NoContentResult`

Happy to tweak this too if there's a way to handle it differently.

mapleleafjack (approved):

PR #537, opened 2021-11-29
SCT-1538 add cp episodes support

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1538

The initial implementation for CP didn't support episodes within the CP status. This is not inline with the business requirements, so have to be changed. 

Treat CP exactly the same way as LAC. Only difference being that LAC always has two answers in a group whereas CP only has one.

I have broken down the test classes to separate files and namespaces to make them easier to follow. It looks like a lot of new code, but most of it is just things moving to new files.

All common tests are now in the original use case and gateway test files and then type specific tests are in their own files. It looks like there's some repetition in the tests between types, but that's on purpose so we can deal with any possible further business rules changes for each type independently.

#### _Checklist_

Test togeter with front end changes to ensure nothing has gone out of sync in the process.


mapleleafjack (approved):
Looks good to me! 🚀

PR #538, opened 2021-11-30
SCT-1346- Add deleted records filter to get resident cases end point

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1543

We have implemented deleted submissions filter to form submissions end point, but that doesn't cover the cases end point that's used by the front end to populate person's timeline. 

1. Add `include_deleted_records` flag support to `GET` `/cases/` end point
2. Update the `CaseRecordsUseCase` to not return deleted records by default
3. Add `DeletedRecordsCount` property to the return object, so front end knows there are deleted records for the given filter (typically for particular mosaic_id when used for timeline)

We don't currently support any delete operation against records in `form_data` collection and that's out of scope of this PR. That support has to be developed later if required.

`DeletedRecordsCount` is populated on each call regardless whether deleted records are requested or not. This will cause additional mongo database call, so performance should be monitored closely after deployment. Call is simple with mosaic_id and deleted properties, so existing index  should help with the performance.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Factories/EntityFactoryTests.cs
@@ -565,6 +565,60 @@ namespace SocialCareCaseViewerApi.Tests.V1.Factories
         }
 
         [Test]
+        public void CaseSubmissionToCareCaseDataReturnsDeletedSetToFalseByDefault()
+        {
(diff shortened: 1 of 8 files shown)


rachel-hutchinson (approved):
Approved, this looks good
Discussed with Tuomo some additional performance improvements that will be picked up in a separate PR

PR #542, opened 2021-12-01
SCT-1548 Add include deleted records count flag support

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1548

Currently form submissions end point returns the deleted records count on every call. After reviewing the potential perormance hit we have agreed to change the integration with the front end so we only fetch this data if it's absolutely required by the UI. Only very limited number of users will ever need that data. 

Add new `includeDeletedRecordsCount` boolean property to `QueryCaseSubmissionsRequest`. Defaults to false.

#### _Checklist_


rachel-hutchinson (approved):
Changes look good, approved

PR #544, opened 2021-12-01
SCT-1548 add deleted records count filter to cases endpoint

LBHTKarki:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1548

Current deleted records count is always returned by the get cases end point. Thus causes unnecessary MongoDB calls and therefore affects performance.

Update cases end point not to return the deleted cases count unless specifically requested by setting the `include_deleted_records_count` flag to true.

#### _Checklist_


rachel-hutchinson (approved):
Approved, changes look good

mapleleafjack (approved):

PR #563, opened 2022-01-11
(SCT-1062) Implement historical data features from platorm api

LBHTKarki:
## Link to JIRA ticket
[SCT-1062](https://hackney.atlassian.net/browse/SCT-1062)

We have a separate platform API for providing access to read only data restored from Mosaic backups. Currently it serves historical case notes and visits data. The initial idea of this kind of platform API doesn't really make sense anymore, so it would be best to streamline the functionality so this data is provided directly by the service API.

There are also some inefficiences in the way the data is currently fetched, so this gives us an opportunity to refactor the data access and make it more efficient and remove some of the API timeout issues we have been experiencing. 

- Initial implementation of the new historical data gateway that fetches data directly from the historical data database rather than going through the platform API
-  Add Docker setup for the new database to enable testing locally and in the pipeline
- Add new database context for the database
- Most of the code was migrated over from the platform API as is, but is has been quite heavily modified to tidy things up
- Tweak some of the boundary object property types to make handling the requests easier
- Where the platform API gateway was used, those integrations have been replaced with the new gateway keeping the domain objects the same to ensure the objects or properties won't change
-  Add test coverage for all new classes and all integration poinst affected
- Add e2e tests to ensure the response objects haven't changed (API end points shouldn't change except for some minor property types mentioned above.)
- Where `DbContextOptions` are used, the database context type has been specified. This is required with EF Core when multiple DB contexts are used
- Update serverless with new connection strings. Connection string for production has been copied from the existing config. Staging one has been added to support staging site.
- Staging database has been added in a separate PR. Database will be empty for now, but it can be populated with seed data later
- Infrastructure objects only include those used by the current service API solution. Everything else have been stripped out to keep things clean
**- The production database is already in place, so any references to schema updates or database setups are purely for staging only and shouldn't be applied to production** . Also this staging database setup is only temporary and will be replaced once we get to migrate the staging resources to a new account

There's also unrelated update for `AddMissingUPRNsTests.cs` class to make the test more robust. 

#### _Checklist_

- Improve the performance further by introducing paging to the gateway methods
- Migrate historical data to service database to simplify data access further (optional)
- Decommission the platform API and remove all resources related to it (except for the database)


mapleleafjack (approved):
A bit of a lengthy PR but looks definitely worth it!
Nice one!

PR #566, opened 2022-01-13
feat(SCT-1630): Link referral to resident

Renny4Real:
[JIRA](https://hackney.atlassian.net/browse/SCT-1630)

## Describe this PR

A client on a referral needs to be matched to a person in the system. A client on a referral is saved as a separate entity in the database called Mash resident. **Note: this is not the same as a Person in the social care system.**

 During the MASH workflow, a worker needs to check that client on a referral already exists as a person in the system or if they need to be added as a new person to the social care system. i.e link the referral to the saved person(s) in the service's DB.

This introduces a new endpoint that updates the `sccv_residents_id` column for a Mash resident record and sets the value to the ID of the matching person in the database.

#### _Checklist_

- [X] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature

Test this works in staging


SocialCareCaseViewerApi.Tests/V1/IntegrationTests/MASH/LinkClientInReferralToResidentInSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            var response = await Client.PatchAsync(patchUri, requestContent).ConfigureAwait(true);
+
+            response.StatusCode.Should().Be(400);
+        }
+
+        [Test]
+        public async Task WhenTheRequestContainsAnIncorrectPersonIdReturnsRelevantErrorMessage()
+        {
+            var incorrectPersonId = _existingDbPerson.Id + 20;
```
martapederiva, line 70: would the test fail should a person with id +20 actually exist in the db? not sure if it make sense, but would using a negative number id (eg. -400) work for this purpose? happy to chat about it

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/MASH/LinkClientInReferralToResidentInSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+            var patchResponse = JsonConvert.DeserializeObject<string>(content);
+
+            patchResponse.Should().Be($"Person with id {incorrectPersonId} not found");
+        }
+
+
+        [Test]
+        public async Task WhenTheRequestContainsAnIncorrectMashResidentIdReturnsRelevantErrorMessage()
+        {
+            var incorrectMashResidentId = _unlinkedResident.Id + 20;
```
martapederiva, line 89: see above

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/MASH/LinkClientInReferralToResidentInSystem.cs
Renny4Real: Hi, so this should take the id of the existing person that we set up in the setup stage and add 20 to it. In theory, that ID would not exist (since we've only setup one person in the DB ) and the test database should be wiped before every test. 20 was randomly selected and wasn't sure if there could be issues with possible negative values which is why I decided to add rather than subtract.

martapederiva: I  forgot that we're not connecting to the actual db for this 🤦 
makes sense, thank you for explaining 🙇‍♀️

martapederiva (approved):
👍 👍 👍

SocialCareCaseViewerApi.Tests/V1/IntegrationTests/MASH/LinkClientInReferralToResidentInSystem.cs
Renny4Real: The reason I didn't want to randomly generate the ID or give a set value was to avoid the rare chance that the ID we want to test with by coincidence, matched the ID of the person we setup in the database. So by using the existing person's ID as a reference, we should never have the same ID appear.

Renny4Real: For the integration tests, we should be connecting to the test database in docker but our integration test setup has been designed in way to make sure that we wipe the test database before each integration test is run.

PR #567, opened 2022-01-13
(SCT-1635) Configuration update for historical data access

LBHTKarki:
## Link to JIRA ticket
[SCT-1635](https://hackney.atlassian.net/browse/SCT-1635)

In preparation for the new historical data gateway deployment we need to ensure the service api has been configured correctly so it can access the historical data database

- Add new `HISTORICAL_DATA_CONNECTION_STRING` environment variable to serverless config for the database connection string. Added for both staging and production environments
- Update Docker setup to include the new test database
- Add new security groups configuration for the Lambda to enable database connection
- Add new temporary end point to health check controller, so the connection can be verified safely
- Add integration tests to verify the env var and connection config

Please note the way the connection is setup in the controller is kept simple intentionally. It is not using proper database context or any other dependencies, but instead it opens a standalone connection, runs a simple query to verify the connection and then disposes it. This ensures any potential infrastructure config issues won't affect the API as a whole and that the new database connection can be tested safely in isolation. The end point can be removed after testing.

#### _Checklist_

After succesfully testing the connection on production, the [PR#563](https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/563) can be merged and deployed.


mapleleafjack (approved):
Nice!

PR #569, opened 2022-01-17
Unlink resident from referral

moarpheus:
## Link to JIRA ticket

[JIRA ticket.](https://hackney.atlassian.net/browse/SCT-1633)

During the MASH workflow, a worker needs to be able to un-assign person from Mash resident. i.e unlink the referral from the saved person(s) in the service's DB.

- Moved error testing cases linking person to mash resident from Integration test to the controller.
- Added condition inside MashRefferalGateway update method to check for presence of "UNLINK-PERSON" update type + with tests and refactored helper classes to support that.
- Integration test to check the happy path.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Controllers/MashResident/UpdateMashResidentTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Controllers/MashResident/UpdateMashResidentTests.cs
@@ -28,7 +28,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Controllers.MashResident
         }
 
         [Test]
-        public void WhenSuccessfullyUpdateActionReturns204AndUpdatedMashResident()
+        public void WhenSuccessfullyUpdateActionReturns200AndUpdatedMashResident()
         {
             var matchingPersonId = _faker.Random.Long(3, 4);
             var request = TestHelpers.CreateMashResidentUpdateRequest(matchingPersonId);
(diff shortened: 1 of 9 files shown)


SocialCareCaseViewerApi.Tests/V1/Controllers/MashResident/UpdateMashResidentTests.cs
```diff
@@ -44,5 +45,37 @@ public void WhenSuccessfullyUpdateActionReturns204AndUpdatedMashResident()
             response?.StatusCode.Should().Be(200);
             response?.Value.Should().BeEquivalentTo(updatedMashResident);
         }
+
+        [Test]
+        public void WhenMashResidentUseCaseUpdateMashResidentThrowsPersonNotFoundExceptionReturnsBadRequest()
+        {
+            var matchingPersonId = _faker.Random.Long(3, 4);
```
martapederiva, line 52: As I've recently been reminded by @Renny4Real (after a comment where I suggested a similar solution with a random number), there is a (very small) chance the ID we want to test with might coincidentally match the ID of the person we setup in the database for the test. So maybe adding a number to the ID itself (as was done in the integration tests that have now moved here) might exclude that chance..?

moarpheus: Long - Generate a random long between **MinValue** and **MaxValue**. From the Bogus source [here](https://github.com/bchavez/Bogus).

`_fakeMashResidentId = _faker.Random.Long(1, 2);`

It can become either 1 or 2

`matchingPersonId = _faker.Random.Long(3, 4);`

It can become either 3 or 4.

Unless Faker will behave unexpectedly there is 0 chance of accidental match.

martapederiva (approved):
LGTM

PR #578, opened 2022-01-25
(SCT-1652) Remove env vars for platform api

LBHTKarki:
## Link to JIRA ticket

[SCT-1652](https://hackney.atlassian.net/browse/SCT-1652)

We have decommissioned two APIs, but we still have environment variables set for those.

Tidy up the serverless config by removing environment variables for decommissioned APIs.

#### _Checklist_


martapederiva (approved):
🧹

PR #589, opened 2022-02-21
Pinned at for case notes

moarpheus:
## Link to JIRA ticket

[JIRA ticket](https://hackney.atlassian.net/browse/SCT-1702).

Adding “pinned” functionality to case notes.

- adding a pinned_at: Date column to case note submissions that can be exposed and updated via the api (/cases endpoint)

- floating pinned notes to the top of any responses for the /cases endpoint, so they are returned first.

Introduced PinnedAt property to the cases, API request and response + ensuring the pinned cases appear before everything else in the list.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
+++ b/SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
@@ -222,6 +222,7 @@ namespace SocialCareCaseViewerApi.V1.Factories
                     FormId = caseSubmission.FormId,
                     CreatedBy = new Worker { Email = caseSubmission.CreatedBy.Email, },
                     CreatedAt = caseSubmission.CreatedAt,
+                    PinnedAt = caseSubmission.PinnedAt,
                     Residents = caseSubmission.Residents.Select(r => new Person
                     {
                         Id = r.Id,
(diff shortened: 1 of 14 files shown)


SocialCareCaseViewerApi/V1/UseCase/FormSubmissionsUseCase.cs
```diff
@@ -219,6 +220,11 @@ public CaseSubmissionResponse ExecuteUpdateSubmission(string submissionId, Updat
             UpdateSubmissionState(updatedSubmission, request, worker);
             UpdateResidents(updatedSubmission, request);
 
+            if (request.PinnedAt != null)
```
jhackett1, line 223: so, we just need to pass `?pinnedAt` into the request url with any value, right?

SocialCareCaseViewerApi/V1/Factories/EntityFactory.cs
```diff
@@ -270,7 +271,8 @@ public static LastUpdated ToDomain(this dbLastUpdated update)
                 LastEdited = caseSubmission.EditHistory.Last().EditTime,
                 CompletedSteps = caseSubmission.FormAnswers.Count,
                 Deleted = caseSubmission.Deleted ?? false,
-                DeletionDetails = includeDeletionDetails ? caseSubmission.DeletionDetails : null
+                DeletionDetails = includeDeletionDetails ? caseSubmission.DeletionDetails : null,
+                PinnedAt = caseSubmission.PinnedAt
```
jhackett1, line 275: is pinned timedate returned in specific GETs of a particular case, as well as the list GET?

jhackett1 (approved):

PR #591, opened 2022-02-25
Toggle pinned cases

moarpheus:
## Link to JIRA ticket

[JIRA ticket.](https://hackney.atlassian.net/browse/SCT-1727)

The pinned sort should only be applied if we pass a pinned_first query parameter.

Toggle in the code to switch to pinned cases when pinned_first is set to true.

#### _Checklist_


martapederiva (approved):
LGTM 👍

PR #592, opened 2022-03-01
Allocate team to resident

moarpheus:
## Link to JIRA ticket

[JIRA ticket.](https://hackney.atlassian.net/browse/SCT-1719)

In order to surface team allocations, we need to allow for teams to be allocated to residents. This will allow us to continue to develop further team allocations iterations. 

Added controller, database gateway, use case code and tests, additional table in DB + additional entry for exclusion list for cases

#### _Checklist_

Staging needs manually migration run.


SocialCareCaseViewerApi/V1/Controllers/ResidentController.cs
```diff
@@ -39,6 +39,39 @@ public IActionResult ListContacts([FromQuery] ResidentQueryParam rqp, int? curso
...
         }
 
+        /// <summary>
+        /// Creates a new Allocation Resident/Team
+        /// </summary>
+        /// <response code="204">Records successfully inserted</response>
+        /// <response code="400">One or more request parameters are invalid or missing</response>
+        [ProducesResponseType(typeof(CreateAllocationResponse), StatusCodes.Status201Created)]
+        [HttpPost]
+        [Route("{residentid}/allocateteam")]
```
acodeninja, line 49: Just a thought, does this route mean there will be an allocateworker endpoint? This doesn't seem quite right to me, my thought was that allocating a worker was effectively a patch request to update the worker that is part of the allocation? 

My thought was to just have `{residentid}/allocate` and allow it to accept a post, then `{residentid}/allocate/{allocationid}` to accept a patch request to update the worker assigned.

Then to action a deallocation I was thinking a delete request to the above route, which would then soft delete the allocation.

Happy to be challenged on this though.

moarpheus: Makes sense, we will amend today!

PR #594, opened 2022-03-08
Circular reference

moarpheus:
## Link to JIRA ticket

Add a link to the JIRA ticket that the changes in this PR describe.

Circular references in the related objects in the MongoDB.

Added test and code to sanitise all possible related objects before storing it in the MongoDB.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/UseCase/FormSubmissionUseCaseTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/UseCase/FormSubmissionUseCaseTests.cs
@@ -1258,7 +1258,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.UseCase
             submission.SubmissionState = SubmissionState.InProgress;
             submission.FormAnswers = new Dictionary<string, string>();
             submission.EditHistory =
-                new List<EditHistory<Worker>> {new EditHistory<Worker> {Worker = workers.First(), EditTime = DateTime.Now}};
+                new List<EditHistory<Worker>> { new EditHistory<Worker> { Worker = workers.First(), EditTime = DateTime.Now } };
 
             var (caseSubmissionResponse, caseSubmissionForPost) = _formSubmissionsUseCase.ExecutePost(request);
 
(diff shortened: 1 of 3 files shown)


SocialCareCaseViewerApi.Tests/V1/UseCase/FormSubmissionUseCaseTests.cs
```diff
@@ -1174,5 +1174,80 @@ public void ExcecuteDeleteCallsMongoDbGaetwayWhenTheSubmissionTypeIsSupported(st
...
+            person.TechUse?.Add(techUse);
+            person.Disability?.Add(disability);
+            person.Emails?.Add(emailAddress);
+            person.LastUpdated?.Add(lastUpdated);
+
+            var residents = new List<Person> { person };
+            var request = TestHelpers.CreateCaseSubmissionRequest();
+            var workers = new List<Worker> { TestHelpers.CreateWorker() };
+
+            _mockDatabaseGateway.Setup(x => x.GetWorkerByEmail(request.CreatedBy)).Returns(workers[0]);
```
LBHTKarki, line 1222: .First() could be used instead of workers[0] for readability

SocialCareCaseViewerApi.Tests/V1/UseCase/FormSubmissionUseCaseTests.cs
```diff
@@ -1174,5 +1174,80 @@ public void ExcecuteDeleteCallsMongoDbGaetwayWhenTheSubmissionTypeIsSupported(st
...
+            _mockDatabaseGateway.Setup(x => x.GetPersonDetailsById(request.ResidentId)).Returns(residents[0]);
+            _mockMongoGateway.Setup(x => x.InsertRecord(It.IsAny<string>(), It.IsAny<CaseSubmission>()));
```
LBHTKarki, line 1224: Because the circular reference issue occurs at this stage I think it would be better to test the shape of the object passed to the`InsertRecord` method rather than testing what comes back from the use case. This way you would be confident that the circular references have been removed from the object to be inserted. The use case has a lot of logic that alters the shape of the object before it's returned, so targeting the input object would be safest option I think.

LBHTKarki (changes requested):

SocialCareCaseViewerApi.Tests/V1/UseCase/FormSubmissionUseCaseTests.cs
moarpheus: Refactored to check which objects is being inserted

LBHTKarki (approved):

PR #603, opened 2022-03-16
fix(SCT-1756): Correction as name is not set when first and last are

acodeninja:
## Link to JIRA ticket

[Add a link to the JIRA ticket that the changes in this PR describe.](https://hackney.atlassian.net/browse/SCT-1756)

Made a small mistake where first and last name are only used if name is set, obviously this is wrong.

Allow first and last name to be concatenated when name is not set.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi/V1/Gateways/SearchGateway.cs
+++ b/SocialCareCaseViewerApi/V1/Gateways/SearchGateway.cs
@@ -63,17 +63,14 @@ namespace SocialCareCaseViewerApi.V1.Gateways
             var name = request.Name == null ? "" : $"{request.Name}";
             var cursor = request.Cursor ?? 0;
 
-            if (string.IsNullOrEmpty(request.Name))
+            if (!string.IsNullOrEmpty(request.FirstName))
             {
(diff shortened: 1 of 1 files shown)


SocialCareCaseViewerApi/V1/Gateways/SearchGateway.cs
```diff
@@ -63,17 +63,14 @@ private List<SearchResult> FilterByNameAgainstAllRecords(PersonSearchRequest req
             var name = request.Name == null ? "" : $"{request.Name}";
```
LBHTKarki, line 63: This will potentially concatenate all three values together, but I suppose client can manage that to ensure the query is sensible

LBHTKarki (approved):

SocialCareCaseViewerApi/V1/Gateways/SearchGateway.cs
acodeninja: It will yes, this is part of the first bit, expansion. A follow up commit will remove the name from the request param.

PR #608, opened 2022-03-17
Add new fields to the allocations table

moarpheus:
## Link to JIRA ticket

[JIRA ticket.](https://hackney.atlassian.net/browse/SCT-1758)

We need to add the new fields to the allocations in order to be able to save new information about allocations.

Added fields to the table and infra as well as added those fields in request and created the missing validations for it. Since we are validating presence of certain fields in the request those must be nullable.

#### _Checklist_

Run manual DB migration for the new fields


moarpheus:
Patching new fields will be done in the next PR.

PR #614, opened 2022-03-18
Improved integration test setup

LBHTKarki:

Our current end to end test setup doesn't reflect the production setup closely enough, so we have some conflicts between how code works on production and what the tests are telling us.

- Update `MockWebApplicationFactory` class to use the same lifetime for the `DatabaseContext` as on production. The current implementation of Singleton service in tests means the behaviour of the application is different from production. Service is now registered using `AddDbContext` method which defaults to scoped lifetime. This matches the lifetime configuration on production `Startup`
- Some tests have been tweaked to allow minor time differences in timestamp comparisons now that the values are correctly persisted to database during testing. Some precision is lost when the data is saved to SQL and that is expected behaviour
-  `IntegrationTestSetup` class has been simplified and updated to use the newly scoped `DatabaseContext` from the `MockWebApplicationFactory`
- `UpdateWorkerWithNewTeamReturnsTheOnlyTheUpdatedTeam` test has been updated to reflect the current functionality of the moving worker to a new team feature. Please note the bug hasn't been fixed, so this is there just to verify the current production behaviour. The feature will need to be updated anyway, so this test can be updated to reflect the new behaviour once completed
- `UpdateWorkerWithNewTeamUpdatesAnyAllocationsAssociated` test has been updated to reflect the current buggy behaviour of the code when worker is moved to a new team (related to the issue above)

It is a bit confusing how the schema file is handled in test setups and initially when test db Docker container is launched. This could be improved further, but since it has no direct connection to the issue this PR is resolving it can be left for later.

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Data/PostgreSQL/AddMissingUPRNsTests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Data/PostgreSQL/AddMissingUPRNsTests.cs
@@ -102,7 +102,7 @@ namespace SocialCareCaseViewerApi.Tests.V1.Data.PostgreSQL
 
             var updatedAddress = DatabaseContext.Addresses.FirstOrDefault();
 
-            updatedAddress.LastModifiedAt.Should().BeCloseTo(DateTime.Now, precision: 5000);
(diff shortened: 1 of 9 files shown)


moarpheus:
Looks good to me!

PR #620, opened 2022-03-24
Update historical data handling when moving workers

LBHTKarki:

The first implementation of new worker move feature wasn't able to handle historical records correctly.

Update the logic for historical records that have none of the new column values so they can be updated correctly and get inline with the new data structure.

#### _Checklist_


martapederiva (approved):
🔥

PR #625, opened 2022-03-28
No team worker bug

moarpheus:

Team without ID is created on deallocation when allocation only for workers exists

Condition check and test

#### _Checklist_


acodeninja (approved):
Great work, thank you!

PR #633, opened 2022-03-29
wraped up deallcation for both team and worker

moarpheus:

We need to be able to deallocate both team and worker allocations via single call.

Adjusted request and code to implement the functionality + test

#### _Checklist_


acodeninja (approved):
🥇

PR #641, opened 2022-04-06
Insert statements for ILDS

martapederiva:
## Link to JIRA ticket

https://hackney.atlassian.net/browse/SCT-1806

Creating the new ILDS teams structure

Insert statements

Run select query to check insert ran smoothly.


mapleleafjack (approved):
🚀

PR #647, opened 2022-04-13
Add test for sorting by date added when filtered by worker id

LBHTKarki:
## Link to JIRA ticket

SCT-1815

Issue was reported saying that when fetching allocations filtered by `worker_id` and setting the `sort_by` parameter to `date_added` the results weren't coming back in the correct order.

Added a test for the above scenario.

#### _Checklist_

Test is passing, so investigate further where the issue is since it looks like it's outside the API.


martapederiva (approved):
🔥

PR #653, opened 2022-04-20
Sct 1726 team allocations import

LBHTKarki:
## Link to JIRA ticket

[SCT-1726](https://hackney.atlassian.net/browse/SCT-1726)

We need to be able to import team allocation data from a spreadsheet provided by the service.

Add SQL scripts for:

1. Importing team allocations i.e. waiting list records. Data set must be pre-validated and filtered to only include correct records i.e. items that only have team id and not worker id
2. Updating person review date values based on the data set provided by the service. Data set must only include items that have review date set. Script won't update existing values, it only adds new ones. This behaviour can be reviewed if necessary.

#### _Checklist_

Get the spreadsheet from the service and start the data cleansing/validation process


LBHTKarki:
This process can now be simplified. I'll create a new PR later

PR #662, opened 2022-04-26
SCT-1832 Get the most recent address first.

moarpheus:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/SCT-1832)

We’ve come across a bug where people end up with two addresses with ‘is_display_address’ = 'Y'

This causes strange frontend behaviour when updating a resident's address.

Changed the order of addresses when we call get resident endpoint and same when we update the address.

Updating address will do following:

- Set previous most recent address to N
- Create new address entry related to the person id

#### _Checklist_

### DIFF
--- a/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
+++ b/SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
@@ -2331,6 +2331,49 @@ namespace SocialCareCaseViewerApi.Tests.V1.Gateways
         }
 
         [Test]
+        public void UpdatePersonsAddressUpdatesTheMostRecentAddressToDisplay()
+        {
+            var first = DateTime.Today;
+            var second = DateTime.Today.AddDays(-1);
+            var last = DateTime.Today.AddDays(-2);
+
+            var person = DatabaseGatewayHelper.CreatePersonDatabaseEntity();
+            person.Addresses = null;
+            var firstAddress = DatabaseGatewayHelper.CreateAddressDatabaseEntity(person.Id, startDate: first, isDisplayAddress: "Y");
+            var secondAddress = DatabaseGatewayHelper.CreateAddressDatabaseEntity(person.Id, startDate: second, isDisplayAddress: "Y");
+            var lastAddress = DatabaseGatewayHelper.CreateAddressDatabaseEntity(person.Id, startDate: last, isDisplayAddress: "Y");
+
+            DatabaseContext.Persons.Add(person);
+            DatabaseContext.Addresses.Add(firstAddress);
+            DatabaseContext.Addresses.Add(secondAddress);
+            DatabaseContext.Addresses.Add(lastAddress);
+            DatabaseContext.SaveChanges();
+
+            var personBeforeUpdate = _classUnderTest.GetPersonDetailsById(person.Id);
+
+            personBeforeUpdate.Addresses.Count.Should().Equals(3);
+            personBeforeUpdate.Addresses.First().StartDate.Should().Equals(first);
+            personBeforeUpdate.Addresses.First().IsDisplayAddress.Should().Equals("Y");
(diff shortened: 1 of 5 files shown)


martapederiva (approved):

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -2330,6 +2330,49 @@ public void UpdatePersonSetsCorrectValuesForPersonEntity()
...
+            var secondAddress = DatabaseGatewayHelper.CreateAddressDatabaseEntity(person.Id, startDate: second, isDisplayAddress: "Y");
+            var lastAddress = DatabaseGatewayHelper.CreateAddressDatabaseEntity(person.Id, startDate: last, isDisplayAddress: "Y");
+
+            DatabaseContext.Persons.Add(person);
+            DatabaseContext.Addresses.Add(firstAddress);
+            DatabaseContext.Addresses.Add(secondAddress);
+            DatabaseContext.Addresses.Add(lastAddress);
+            DatabaseContext.SaveChanges();
+
+            var personBeforeUpdate = _classUnderTest.GetPersonDetailsById(person.Id);
```
LBHTKarki, line 2352: If the implementation of this method changes then this test will potentially be affected. Would it be safer to use the DatabaseContext directly when fetching records?

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -2330,6 +2330,49 @@ public void UpdatePersonSetsCorrectValuesForPersonEntity()
...
+
+            var updatedPerson = _classUnderTest.GetPersonDetailsById(person.Id);
+
+            updatedPerson.Addresses.Count.Should().Equals(4);
+            updatedPerson.Addresses.First().StartDate.Should().Equals(first);
+            updatedPerson.Addresses.First().PostCode.Should().Equals(firstAddress.PostCode);
+            updatedPerson.Addresses.First().EndDate.Should().NotBeNull();
+            updatedPerson.Addresses.First().IsDisplayAddress.Should().Equals("N");
+
+            updatedPerson.Addresses.Last().IsDisplayAddress.Should().Equals("Y");
```
LBHTKarki, line 2371: would it be safer not to rely on the order and identify the new entry by its id? That would ensure the newly added address really is the one that's been set to be the display address? Also might be better to use the db context here as well when fetching the records.

SocialCareCaseViewerApi.Tests/V1/Gateways/DatabaseGateway.Tests.cs
```diff
@@ -2330,6 +2330,47 @@ public void UpdatePersonSetsCorrectValuesForPersonEntity()
...
+            personBeforeUpdate.Addresses.Count.Should().Equals(3);
+            addressBeforeUpdate.IsDisplayAddress.Should().Equals("Y");
+
+            var request = GetValidUpdatePersonRequest(person.Id);
+
+            _classUnderTest.UpdatePerson(request);
+
+            var updatedPerson = DatabaseContext.Persons.Find(person.Id);
+            var previousAddress = DatabaseContext.Addresses.FirstOrDefault(x => x.PostCode == firstAddress.PostCode && x.StartDate == first && x.PersonId == person.Id);
+            var newAddress = DatabaseContext.Addresses.FirstOrDefault(x => x.PostCode == request.Address.Postcode && x.PersonId == person.Id);
```
LBHTKarki, line 2364: If you have the original IDs in `existingAddresses` list, you can grab the new address by id by excluding the old ones, like `var newAddress = DatabaseContext.Addresses.FirstOrDefault(x => !existingAddresses.Select(x => x.AddressId).Contains(x.AddressId));`

That should make everything ID based so the values don't matter at all.

LBHTKarki (approved):

